PR #393, opened 2018-02-06
Issue #1489: Make our TLS fingerprint more like Golang TLS

hchood:
TLS fingerprinting by Sergey & Eric at UColorado shows that there are a few differences between Lantern’s fingerprint and the fingerprint generated by the standard Go TLS library.  

These include:

1. **Ordering of cipher suites.**  

While we rely on Go TLS's default ordering, their Lantern fingerprint prioritized CHACHA over AES, while their Golang fingerprint showed AES being prioritized.  This could be due to differing use of desktop vs mobile devices in their sample, however, as Go's TLS library does prioritize CHACHA due to hardware limitations in some cases:

> this ordering could be generated by standard Go on devices without hardware support for AES-GCM: https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L914

Indeed, when I followed up for clarification they said they had only tested Lantern on mobile, not desktop.

2. **SNI header**.  

Our fingerprint did not contain an SNI header; the Golang fingerprint did.  For now, I'm leaving the SNI header blank by default, but allowing it to be set via the proxy config.  

[*TODO*: Determine whether we should actually be setting SNI even if it is not set in the proxy config -- see discussion in slack.]

3. **SessionTicket extension**.  

Our fingerprint did contain an SNI header; the Golang fingerprint did not, as you must set the client session cache to enable session tickets:

https://github.com/golang/go/blob/e4bde0510465eecd4c8a8293418b1cbed1e0e623/src/crypto/tls/handshake_client.go#L115-L122

However, I'm not going to do anything on this point now unless instructed otherwise given there are performance implications for not using session tickets.


myleshorton (commented):
I’m still a little confused on the cipher suites @hchood. Like if we never specify the cipher suites it will always just use the default on both mobile and desktop right? Seems like we can’t do better than that to me.

PR #1225, opened 2022-04-27
Russia live test

anacrolix:
This should be squashed to begin the Russia live-test.

@soltzen @oxtoacart @myleshorton do you want to check this over? I've pieced this together from our previous commits on this, and the Live-test config updates that Alec provided. To test I think we merge this, and then revert the merge to flip it back.

https://github.com/getlantern/lantern-internal/issues/5461

### DIFF
--- a/embeddedconfig/global.yaml.tmpl
+++ b/embeddedconfig/global.yaml.tmpl
@@ -203,7 +203,7 @@ featureoptions:
     # These are for compatibility with clients that don't load all options per-country.
     replicarustdefaultendpoint: &FrankfurtReplicaRust http://replica-search-aws.lantern.io/
     replicarustendpoints:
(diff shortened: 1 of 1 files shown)


soltzen (approved):
LGTM.

I suggest squashing all commits into one, so that when we wanna `git revert` the PR (after 1 hour of testing), it'll be much easier.

Furthermore, it occurs to me that there are some lantern-desktop changes that are required **before** publishing, namely the pathing issue: https://github.com/getlantern/replica/pull/66

Without those changes, lantern-desktop won't play any videos. This means old clients won't be able to use our RU infrastructure.

Maybe we should publish a new version of lantern-desktop first, get a few people on it, and then run the test. Without it, Replica won't work. Lemme know if I misunderstood something @anacrolix

PR #1252, opened 2022-09-22
Support fetching config in background on iOS

oxtoacart:


myleshorton (changes requested):
I think adhering to the server-side sleep field is pretty vital

myleshorton:
New iOS should cover it

PR #1264, opened 2022-10-19
Add a build flag to turn on config-server proxying

lincolnmantracer:
Config-server has the ability to proxy a config request to lantern-cloud, which is a provisional way to enable lantern-cloud proxies. In order to test this, we need a way for developers to turn on that functionality, and we already use build flags in lantern-desktop for similar purposes.

### DIFF
--- a/common/headers.go
+++ b/common/headers.go
@@ -10,6 +10,12 @@ import (
 	mrand "math/rand"
 )
 
+// EnableLanternCloudReferralHeader is a build flag which enables config
+// proxying to the lantern-cloud API. Proxying works by setting a header. When
+// config-server sees the header, it proxies the config request to the
+// lantern-cloud API as a 'referral'.
+var EnableLanternCloudReferralHeader string
+
 const (
 	AppHeader                           = "X-Lantern-App"
 	VersionHeader                       = "X-Lantern-Version"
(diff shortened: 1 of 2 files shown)


soltzen (approved):
LGTM. I added a tiny amendment to make the flag case-insensitive since I know I'll forget if it should be "TRUE" or "true" or "1". If you disagree, just drop the commit. Squash it otherwise.

Also, I'd love a makefile target for lantern-desktop (and optionally android, but not so important) that allows you to target lantern-cloud by setting the flag for you.

Something like this [here](https://github.com/getlantern/lantern-desktop/blob/b04942d185917d9fcf2b0ff6353cbeae5bda5c15/Makefile#L162):

```
ifdef LANTERN_CLOUD
EXTRA_LDFLAGS += -X github.com/getlantern/flashlight/common.EnableLanternCloudReferralHeader=true
endif
```

And then later:

```
# This runs a development build for lantern against lantern-cloud
.PHONY: lantern-devel-with-lantern-cloud
lantern-devel-with-lantern-cloud: export LANTERN_CLOUD = true
lantern-devel-with-lantern-cloud: export DISABLE_PORT_RANDOMIZATION = true
lantern-devel-with-lantern-cloud: export ENVIRONMENT = development
lantern-devel-with-lantern-cloud: export GO_BUILD_FLAGS += $(DISABLE_OPTIMIZATION)
lantern-devel-with-lantern-cloud: export BUILD_RACE = -race
lantern-devel-with-lantern-cloud:
	@make app

```

PR #1298, opened 2023-04-10
Revamp integration test package

soltzen:
Related to https://github.com/getlantern/lantern-internal/issues/5972
Depends on https://github.com/getlantern/http-proxy-lantern/pull/553

Reworks the existing (non-functional) integration testing package: it now runs a local Flashlight instance against a local http-proxy-lantern instance.

I've written a few tests already (for http, https and shadowsocks protocols). More cases to come.

Furthermore, I've made the logging for that package go to a file since Flashlight is notoriously verbose. I'm planning to make a TUI (text user interface) and have the tool be usable in CI as well, but this is a first stab only


hwh33 (commented):
It seems like the two main outputs of this are:
1. Split out a new package of local-proxy implementations.  This is now called `proxyimpl`.
2. Define a command for running integration tests using a local version of http-proxy-lantern for the remote proxy.

On point 1, I think the new interface boundary needs a bit of work.  I left some comments to this effect.

On point 2, I think this is a positive change, but it adds a lot of overhead in terms of code and project complexity.

What if you wrote the tests as actual Go tests? So the `integrationtest` package would define facilities for setting up a local test proxy, config pairs, etc. Then you could define `TestHTTP`, `TestHTTPS`, `TestShadowsocks`, etc. You'd avoid all of the overhead and complexity of defining your own test runner. Most importantly, the user wouldn't need to learn anything new when running tests or creating new ones. Additionally, the tests would run as part of `go test ./...` from the top-level directory in flashlight, without the need for special handling of `integrationtest` and parsing of its results file.

I know you've put a lot of work into this.  However, I'd like to see this land in the simplest and most usable form.  I think defining our own test runner is unnecessary when we have a robust one available to us in the Go standard library.

To be clear, I think both (1) and (2) are hugely beneficial!  Thanks for taking on this effort.

teamlantern:
This PR has been inactive for 1054 days.

Waiting on: none

It will be closed automatically in 7 days if no action is taken.

PR #1308, opened 2023-06-17
Send traces directly from bypass

hwh33:
This PR:
- disables traces sent by bypass via the chained dialer
- adds traces sent directly by bypass
- distinguishes traces sent via domain fronting or normal proxying

This resolves https://github.com/getlantern/engineering/issues/202.

Depends on https://github.com/getlantern/lantern-cloud/pull/277


hwh33:
Forgot about this PR for an embarrassing amount of time 😅 

After taking another look at Ox's suggestions here, I think he's right that it would be better to just add an op for Bypass dials.  I've done this in a new PR here: https://github.com/getlantern/flashlight/pull/1336

PR #1329, opened 2023-10-31
Bump to latest http-proxy-lantern

myleshorton:


myleshorton:
This came up in slack, but the recent compilation issue was the result of [this](https://github.com/anacrolix/torrent/issues/859).

Ideally to me we'd update the direct dependencies to the correct version and would then bring in all the correct indirect dependencies via that, but in the interest of getting everything back up and working, I'm bringing this in now.

PR #1330, opened 2023-11-07
Fully remove borda from flashlight

myleshorton:
This fully strips out borda from flashlight along with the associated dependencies.


go.mod.rej
hwh33, line 1: Looks like this and go.sum.rej snuck in (I think these are from some failed `git` operation).

hwh33 (approved):
Hey sorry @myleshorton, forgot about this one for a minute.

This looks good to me, but there are a couple of informational files from `git` which snuck in and will need to be removed.

PR #1331, opened 2023-11-07
use Go 1.21 and update quic-go version

atavism:
Depends on https://github.com/getlantern/quicwrapper/pull/40 and https://github.com/getlantern/broflake/pull/183/files


myleshorton:
We ended up doing this in another pass -- closing but thanks for giving it a go @atavism!

PR #1335, opened 2023-11-26
Report flashlight library version separately

myleshorton:
We introduced [a change about 6 months ago](https://github.com/getlantern/flashlight/pull/1307) to report versions based on the flashlight version versus the application version. Unfortunately, this meant things such as our client version in BigQuery always being the flashlight version versus the application version, making it difficult to determine the current state of extant versions. 

We also never set nor used the `CompileTimeApplicationVersion` that was still kicking around in the code.

This change does two things:

1. Reports the flashlight version under a separate `X-Lantern-Library-Version` header
2. Restores the reporting of the application version in the `X-Lantern-Version` header


hwh33 (approved):
LGTM.  I would squash the commits since this is one logical change.

hwh33:
Double-checking against the PR description:
> Restores the reporting of the application version in the X-Lantern-Version header

This was changed so that the application version is now reported in the X-Lantern-App-Version header, correct?

hwh33 (approved):
LGTM.  Could you squash before merging?

PR #1337, opened 2023-11-28
Add kernelarch header

myleshorton:
This should allow us to determine if the user's kernel architecture is 32 or 64 bit, which is relevant on mobile in particular where we'd prefer to ditch 32 bit entirely so we can move from netstack to gvisor.

FYI @atavism @jigar-f


myleshorton:
Pulling this in.

PR #1338, opened 2023-12-11
Enabling probes in the US, China, and Iran

myleshorton:


embeddedconfig/global.yaml.tmpl
```diff
@@ -138,10 +138,16 @@ featuresenabled:
     - label: cn-shortcut
       geocountries: cn
       versionconstraints: "!7.4.0"
-  #  - label: ir-desktop-shortcut
-  #    geocountries: ir
-  #    platforms: windows,darwin,linux
+  # - label: ir-desktop-shortcut
+  #   geocountries: ir
+  #   platforms: windows,darwin,linux
   # probeproxies is not enabled anywhere
```
hwh33: This comment is now out-of-date

embeddedconfig/global.yaml.tmpl
```diff
@@ -138,10 +138,16 @@ featuresenabled:
...
+  probeproxies:
+    - label: probeir
+      geocountries: ir
+    - label: probeuscn
+      geocountries: us,cn
```
hwh33: Why are these combined, but the Iran probes are separate?  I don't know much about our proxy probing logic

myleshorton: I think this is just about having separate labels AFIACT, but we don't actually use the labels anywhere on the client side, so I just went ahead and consolidated them.

embeddedconfig/global.yaml.tmpl
myleshorton: Fixed!

myleshorton:
Thanks @hwh33! I think I addressed those comments.

hwh33 (approved):
LGTM

PR #1339, opened 2023-12-12
Notify watchers when geolocation is not from disk

myleshorton:
Starting to cache geolookups on disk 2 years ago meant that we now never notify watchers when we actually learn our geo data, disabling all sorts of things, such as geo-based feature flags, shortcut entirely, etc etc. This should fix it!


myleshorton:
Gonna pull this in.

PR #1340, opened 2023-12-13
Add plans and payment methods to Pro client

atavism:


myleshorton:
I'm a little confused about this one @atavism. Is this because this is currently done in `lantern-client`? Does the pro server actually support these protocol buffers requests and responses?

myleshorton:
Or are you taking advantage of the fact that the proto definition can also support JSON responses?

atavism:
> Is this because this is currently done in lantern-client? Does the pro server actually support these protocol buffers requests and responses?

@myleshorton, we actually have a few different Pro client implementations, and I thought we could update Android and desktop to both use this one. And that's right: the protocol buffer requests and responses are the same that are being used on the pro server. 

> Or are you taking advantage of the fact that the proto definition can also support JSON responses?

Yeah, just being able to generate the structs and JSON from the protocol buffer definitions I feel is really useful as well. If you think this is overkill though, I am fine with removing it!

myleshorton:
Ok LGTM!

PR #1343, opened 2024-01-08
added algeneva support

garmr-ulfr:


go.sum
hwh33, line 1: Needs a `go mod tidy`

chained/algeneva_impl.go
```diff
@@ -0,0 +1,37 @@
...
+		addr:           addr,
+		reportDialCore: reportDialCore,
+	}, nil
+}
+
+func (a *algenevaImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
+	return a.reportDialCore(op, func() (net.Conn, error) {
+		return algeneva.DialContext(ctx, "tcp", a.addr, a.dialerOps)
+	})
+}
```
hwh33: I think this is backwards.  This will still work, but will mean that the dial metrics track the entire round-trip of the `algenava` dial, not just the TCP dial.  I think we want to track the TCP dial, since (a) that reflects the time to open the connection and (b) is more comparable across transports.

However, I just noticed that many other transports take the approach you took here - Shadowsocks for example.  Sorry if you used one of those as an example.  Between the transports we use in practice, there seems to be a 50/50 split between whether we record metrics on the transport-layer (TCP) dial or the circumvention-layer dial.

Going forward, let's track the TCP dial.  Here are a couple of examples of that pattern: (1) [lampshade](https://github.com/getlantern/flashlight/blob/db3236d6286a43fbe205250d29b25b7a2af3e69e/chained/lampshade_impl.go#L97-L105), (2) [tlsmasq](https://github.com/getlantern/flashlight/blob/db3236d6286a43fbe205250d29b25b7a2af3e69e/chained/tlsmasq_impl.go#L141-L143)

garmr-ulfr: > However, I just noticed that many other transports take the approach you took here - Shadowsocks for example. Sorry if you used one of those as an example. Between the transports we use in practice, there seems to be a 50/50 split between whether we record metrics on the transport-layer (TCP) dial or the circumvention-layer dial.

I think it might be because of how the `reportDialCoreFn` is defined. It takes an `*ops.Op` and a function that returns a `net.Conn` and an error, then it itself returns a `net.Conn`. lantern-algeneva doesn't support  wrapping a `net.Conn`, and unfortunately, `net.Dialer` isn't an interface. I think that might be why the other transports also do it this way. I'll have to add support for wrapping a `net.Conn` in lantern-algeneva in order to swap this.

hwh33: Here are a couple examples of two different ways to facilitate this: (1) [wrapping a provided dialer](https://github.com/getlantern/tlsmasq/blob/6e479a593298e24362dd3659c527f0ae75902cc8/tlsmasq.go#L62-L64) or (2) [wrapping an already-established transport connection](https://pkg.go.dev/crypto/tls#Client).

chained/algeneva_impl.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (a *algenevaImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
+	dialerOps := a.dialerOps
+	dialerOps.Dialer = &algenevaDialer{
+		a.reportDialCore,
+		op,
+	}
+
+	return algeneva.DialContext(ctx, "TCP", a.addr, dialerOps)
```
hwh33: This needs to be lowercase: "tcp".  Otherwise, [you get an error](https://go.dev/play/p/QOlbBoxDKA5): `dial TCP: unknown network TCP`.

hwh33 (approved):
One small change needed.  Otherwise, this looks good!

chained/algeneva_impl.go
garmr-ulfr: That's right..oops..

PR #1346, opened 2024-01-11
downgrade to go 1.20 for win7 compatibility

myleshorton:


myleshorton:
OK the issue submission test is failing, but that's unrelated to this change.

PR #1347, opened 2024-01-17
Add tls record fragmentation

myleshorton:


chained/proxy.go
```diff
@@ -324,36 +323,6 @@ func newProxy(name, addr, protocol, network string, s *config.ProxyConfig, uc co
...
-			dfConn, err := p.reportedDial(func(op *ops.Op) (net.Conn, error) { return dial(network, addr) })
-			dfConn, err = overheadWrapper(true)(dfConn, op.FailIf(err))
-			if err == nil {
-				dfConn = idletiming.Conn(dfConn, IdleTimeout, func() {
-					log.Debug("enhttp connection idled")
-				})
-			}
-			return dfConn, err
-		}
-	}
```
myleshorton, line 356: We don't use this anymore

chained/proxy.go
```diff
@@ -114,6 +110,9 @@ func CreateDialer(configDir, name string, s *config.ProxyConfig, uc common.UserC
 		return nil, err
 	}
 	p, err := newProxy(name, addr, transport, network, s, uc)
+	if err != nil {
+		return nil, err
+	}
```
myleshorton, line 117: drive-by

chained/https_impl.go
```diff
@@ -104,7 +109,7 @@ func (impl *httpsImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, er
 func timeoutFor(ctx context.Context) time.Duration {
 	deadline, ok := ctx.Deadline()
 	if ok {
-		return deadline.Sub(time.Now())
+		return time.Until(deadline)
```
myleshorton, line 117: drive-by

chained/proxy.go
hwh33: Wow, good catch!

chained/tlsfrag.go
```diff
@@ -0,0 +1,40 @@
...
+		if err != nil {
+			return conn
+		}
+		return tlsFragConn
+	}
+	return conn
+}
+
+func splitHelloInfo(pc *config.ProxyConfig) (bool, int) {
+	indexStr, ok := pc.PluggableTransportSettings["tlsfrag_split_index"]
```
hwh33: Could you implement this in a way that would support more complex future configurations?  The two I currently have in mind are (1) random within a range and (2) splitting the SNI.

My suggestion would be to rename the option `tlsfrag` and expect a value of the form `frag mechanism, { mechanism-specific config }`.  Example values would be `index:7`, `sni` (randomly within the SNI), `sni:3` (byte 3 of the SNI), `random` (randomly anywhere), `random:5,100` (random within bytes 5 to 100).

As-is, it would be a little awkward to add one of these alternative fragmentation mechanisms in a backwards-compatible way.  For example, if we add a new option `tlsfrag_sni`, what does the client do when both that and `tlsfrag_split_index` are specified?

chained/tlsfrag.go
```diff
@@ -0,0 +1,40 @@
...
+func tlsFragConn(conn net.Conn, pc *config.ProxyConfig) net.Conn {
+	splitHello, index := splitHelloInfo(pc)
+	if splitHello {
+		if _, ok := conn.(*net.TCPConn); !ok {
+			return conn
+		}
+		tlsFragConn, err := tlsfrag.WrapConnFunc(conn.(*net.TCPConn), func(record []byte) (n int) {
+			return index
+		})
+		if err != nil {
```
hwh33, line 22: I feel like we should log here so that we have some indication that fragmentation was not used.

chained/https_impl.go
```diff
@@ -82,6 +85,8 @@ func (impl *httpsImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, er
 			if impl.tlsClientHelloSplitting {
 				tcpConn = hellosplitter.Wrap(tcpConn, splitClientHello)
 			}
+			tcpConn = tlsFragConn(tcpConn, impl.proxyConfig)
```
hwh33: This makes it look like we always fragment.  I think it'd be more readable if the setting was checked explicitly here, like the other settings.  Not a huge deal though.

chained/tlsmasq_impl.go
```diff
@@ -142,6 +148,7 @@ func (impl *tlsMasqImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn,
 	if impl.tlsClientHelloSplitting {
 		tcpConn = hellosplitter.Wrap(tcpConn, splitClientHello)
 	}
+	tcpConn = tlsFragConn(tcpConn, impl.proxyConfig)
```
hwh33: Same comment as on the HTTPS transport:
https://github.com/getlantern/flashlight/pull/1347#discussion_r1456459403

chained/tlsfrag.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/Jigsaw-Code/outline-sdk/transport/tlsfrag"
+	"github.com/getlantern/common/config"
+)
+
+func tlsFragConn(conn net.Conn, pc *config.ProxyConfig) net.Conn {
+	splitHello, index := splitHelloInfo(pc)
+	if splitHello {
+		if _, ok := conn.(*net.TCPConn); !ok {
+			return conn
+		}
```
hwh33: Wait does this work in a production context?  I thought we wrap the transport with our own type.

chained/tlsfrag.go
myleshorton: I feel like you have a clearer picture as to what that should look like than I do. You want to just run with that and implement it on top of this PR?

chained/https_impl.go
myleshorton: Yeah it's almost more like a filter, but tbh I don't love the duplicate logic of the other approach

chained/tlsfrag.go
hwh33: No problem!

859f05b0cd6a0d11ae48bd1d4a554bbf65317dfc

chained/tlsfrag.go
```diff
@@ -0,0 +1,53 @@
...
+	return conn
+}
+
+func makeFragFunc(pc *config.ProxyConfig) (tlsfrag.FragFunc, bool) {
+	fragStr, ok := pc.PluggableTransportSettings["tlsfrag"]
+	if !ok {
+		return nil, false
+	}
+
+	// fragStr should be of the form <frag func> or <frag func>:<func config>.
```
hwh33, line 36: Does this comment make sense to you @myleshorton?  I wasn't sure what would be the most straight-forward in terms of nomenclature or formatting.

hwh33: I added a line below this with an example

chained/tlsfrag.go
myleshorton: Oh very good point....I wonder what it actually is there...will look.

chained/tlsfrag.go
```diff
@@ -75,3 +73,17 @@ func makeFragFunc(pc *config.ProxyConfig) (tlsfrag.FragFunc, bool) {
...
+
+func (sc *streamConn) CloseRead() error {
+	return sc.Close()
+}
+
+func (sc *streamConn) CloseWrite() error {
+	return sc.Close()
+}
+
+var _ transport.StreamConn = (*streamConn)(nil)
```
hwh33, line 117: Nice, I always forget about this trick.

chained/tlsfrag_test.go
```diff
@@ -20,7 +20,7 @@ func TestTLSFragConn(t *testing.T) {
 	resultType := reflect.TypeOf(result)
 	require.Equal(t, "*net.TCPConn", resultType.String())
 
-	pc.PluggableTransportSettings["tlsfrag_split_index"] = "1"
+	pc.PluggableTransportSettings["tlsfrag"] = "index:1"
```
hwh33, line 35: Ah, thank you.  Sorry about that!

chained/tlsfrag.go
```diff
@@ -32,20 +34,41 @@ func makeFragFunc(pc *config.ProxyConfig) (tlsfrag.FragFunc, bool) {
...
 		}
 		return func(_ []byte) int { return index }, true
+	case "regex":
+		regex, err := regexp.Compile(cfg)
+		if err != nil {
+			log.Errorf("tlsfrag: bad regex specifier: %v", err)
+			return nil, false
+		}
+		return func(b []byte) int {
+			loc := regex.FindIndex(b)
```
hwh33: I'm not totally sure that this works.  It sounds like the `regexp` package expects the input to be UTF-8 code-points.  I'm not sure how that would behave with a ClientHello.  I think the text, like the value of the SNI, might be UTF-8 encoded.  But in theory there could be unexpected collisions with the binary data.  I guess that's unlikely?  But technically the behavior would be undefined... or just sort of non-intuitive.  It'd also be difficult to look for stuff like the SNI extension without knowing the SNI ahead of time.  You could look for the expected UTF-8 code-points, but that's pretty awkward IMO.

It's a bummer because specifying a split function as regex would be super powerful.  I thought it was an awesome idea.

hwh33: Oh, what if the regular expression expects to see the ClientHello as a hex-encoded string?  To be concrete, here you would have:
```
return func(b []byte) int {
    loc := regex.FindStringIndex(hex.EncodeToString(b))
    // handle loc == nil
    return loc[0]
}
```

chained/tlsfrag.go
```diff
@@ -32,20 +34,41 @@ func makeFragFunc(pc *config.ProxyConfig) (tlsfrag.FragFunc, bool) {
...
+			if loc == nil {
+				return 0
+			}
+			return loc[0]
+		}, true
+	case "indexOf":
+		return func(b []byte) int {
+			index := bytes.Index(b, []byte(cfg))
```
hwh33: This has the same problem as the regular expression function, but could also be solved by treating the input as a hex-encoded string.

chained/https_impl.go
```diff
@@ -85,7 +88,9 @@ func (impl *httpsImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, er
 			if impl.tlsClientHelloSplitting {
 				tcpConn = hellosplitter.Wrap(tcpConn, splitClientHello)
 			}
-			tcpConn = tlsFragConn(tcpConn, impl.proxyConfig)
+			for _, wrapper := range impl.connWrappers {
+				tcpConn = wrapper(tcpConn, impl.proxyConfig)
+			}
```
hwh33, line 93: Oh this is nice!

myleshorton:
OK I think this is all set @hwh33. I added more tests and a generic net.Conn wrapper func

chained/tlsfrag.go
myleshorton: Interesting yeah I like the hex encoding idea, although I think it would throw the index off? Maybe we could figure out what the index should be though?

At least in that test with the test hello, the straight regex works, but I agree the regex theoretically is looking for UTF-8. I thought it might bork at non-UTF-8 bytes, but that doesn't appear to be true in the test at least.

chained/tlsfrag.go
myleshorton: OK added the hex encoding with just dividing by two to get the index -- good idea!

myleshorton: Oh I ditched the indexOf btw because you can just use regex with a plain string to achieve the same thing (i.e. bytes.Index is the same as regex.FindIndex with a naked string.

myleshorton:
Pulling this in to get a desktop version out. Thanks for the review @hwh33!

PR #1352, opened 2024-01-27
First pass at using multi-arm bandit to select dialers

myleshorton:
This is has been tested extensively running as a user in China with multiple proxies.

~TODO: I still need to add back in the ability to only send broflake specific domains.~


myleshorton:
@hwh33 I think this is actually ready to roll. It's working well for me in tests. Still have to put in a desktop and a mobile PR though.

myleshorton:
OK that's here: https://github.com/getlantern/lantern-desktop/pull/390 for desktop

client/client.go
```diff
@@ -249,7 +250,6 @@ func NewClient(
 	if mitmErr != nil {
 		log.Errorf("Unable to initialize MITM: %v", mitmErr)
 	}
-	client.reportProxyLocationLoop()
```
myleshorton, line 252: We just handle this in our callback now.

client/client.go
```diff
@@ -830,22 +807,27 @@ func (client *Client) ConfigureGoogleAds(opts config.GoogleSearchAdsOptions) {
...
-	}()
-
+	dialer, err := orchestrator.NewWithCallback(dialers, func(d orchestrator.Dialer) {
+		countryCode, country, city := proxyLoc(d)
+		client.statsTracker.SetActiveProxyLocation(
+			city,
+			country,
+			countryCode,
+		)
+		client.statsTracker.SetHasSucceedingProxy(true)
```
myleshorton: This is technically slightly wrong I suppose in that if all of your dialers stop working at some point, the UI won't show that. I'll take a look at solutions.

myleshorton: Fixed!

myleshorton:
If anyone has spare cycles to give this a look, it would be much appreciated!

I think it could help quite a bit. I'll beef up the tests along with the broflake change.

config/config.go
```diff
@@ -110,16 +109,6 @@ type options struct {
 	rt http.RoundTripper
 }
 
-var globalConfigDhtTarget krpc.ID
-
-func init() {
```
reflog, line 115: why was this removed?

client/client.go
```diff
@@ -115,7 +111,8 @@ type Client struct {
 	requestTimeout time.Duration
 
 	// Balanced CONNECT dialers.
-	bal *balancer.Balancer
+	//bal *balancer.Balancer
```
reflog: can probably remove `//bal *balancer.Balancer` and update the comment

chained/tlsmasq_impl.go
```diff
@@ -73,7 +69,7 @@ func newTLSMasqImpl(configDir, name, addr string, pc *config.ProxyConfig, uc com
 		return nil, errors.New("malformed server address: %v", err)
 	}
 
-	proxyTLS, hellos, err := tlsConfigForProxy(ctx, configDir, name, pc, uc)
+	proxyTLS, hellos, err := tlsConfigForProxy(context.Background(), configDir, name, pc, uc)
```
reflog, line 72: why was the timeout removed?

chained/proxy.go
```diff
@@ -71,18 +71,19 @@ type nopCloser struct{}
 
 func (c nopCloser) close() {}
 
-// CreateDialers creates a list of Proxies (balancer.Dialer) with supplied server info.
-func CreateDialers(configDir string, proxies map[string]*config.ProxyConfig, uc common.UserConfig) []balancer.Dialer {
+// CreateDialers creates a list of Proxies (orchestrator.Dialer) with supplied server info.
```
reflog: there's inconsistency in naming. there are several places referencing 'orchestrator' while others just name 'bandit'

chained/https_impl.go
```diff
@@ -31,12 +31,7 @@ type httpsImpl struct {
 
 func newHTTPSImpl(configDir, name, addr string, pc *config.ProxyConfig, uc common.UserConfig, dialCore coreDialer,
 	connWrappers ...connWrapper) (proxyImpl, error) {
-	const timeout = 5 * time.Second
-
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-
-	tlsConfig, hellos, err := tlsConfigForProxy(ctx, configDir, name, pc, uc)
+	tlsConfig, hellos, err := tlsConfigForProxy(context.Background(), configDir, name, pc, uc)
```
reflog, line 34: why was the timeout removed?

bandit/bandit.go
```diff
@@ -0,0 +1,296 @@
...
+	o.onFailure()
+	return nil, &BanditError{
+		attempts: len(o.dialers),
+	}
+}
+
+// Choose a different arm than the one we already have, if possible.
+func (o *BanditDialer) differentArm(existingArm int, numDialers int) int {
+	// We let the bandit choose a new arm 10 times versus just rotating to the next
+	// dialer because we want to the bandit's algorithm for optimizing exploration
```
reflog: this comment seems unclear

bandit/bandit.go
```diff
@@ -0,0 +1,296 @@
...
+		// We don't give any special reward for a successful dial here and just rely on
+		// the normalized raw throughput to determine the reward. This is because the
+		// reward system takes into account how many tries there have been for a given
+		// "arm", so giving a reward here would be double-counting.
+
+		// Tell the dialer to update the bandit with it's throughput after 5 seconds.
+		dt := newDataTrackingConn(conn)
+		time.AfterFunc(secondsForSample*time.Second, func() {
+			speed := normalizeReceiveSpeed(dt.dataRecv)
+			log.Debugf("Dialer %v received %v bytes in %v seconds, normalized speed: %v", dialer.Name(), dt.dataRecv, secondsForSample, speed)
```
reflog: this logging will explode during runtime. maybe keep this commented out?

reflog (changes requested):
awesome stuff. left some notes mostly around naming/comments

myleshorton:
Thanks @reflog will address right away!!

config/config.go
myleshorton: We weren't actually using it, and we just overlooked removing it.

chained/tlsmasq_impl.go
myleshorton: This was the result of the bug I mentioned on slack where all of the hellobrowser-based transports were not picking up the clienthello successfully, so each one would result in a 5 second delay getting connected. If you have 6 proxies, that would be a 30 second delay because it would wait for all of them when creating the dialers map.

reflog (approved):
LGTM, just need to improve test coverage slightly

PR #1354, opened 2024-01-31
Add PaymentRedirect, rename Plan and PaymentMethod

atavism:
...to avoid conflict with internalsdk protobuf messages: https://github.com/getlantern/lantern-client/blob/android-migrate/protos_shared/vpn.proto


pro/client/client.go
```diff
@@ -325,6 +352,8 @@ func (c *Client) execute(user common.UserConfig, method, path string, query url.
 		return err
 	}
 
+	req.Header.Set("Referer", "http://localhost:37457/")
```
atavism, line 355: Since the redirect method is used by payment providers to redirect back to our app's success/fail states - the API expects you to pass 'Referer' header here

PR #1356, opened 2024-02-02
Create dialers in parallel to avoid them blocking first connect

myleshorton:
Currently each dialer, particularly hello-browser-based dialers, can take a long time to connect (5 seconds). We create them serially, so 7 dialers means a 35 second delay on startup. This parallelizes that.


myleshorton:
Pulling this in...

PR #1360, opened 2024-02-28
Improve multi-armed bandit dial timeout handling

myleshorton:
This improves our handling of dial timeouts with multiple dialers for multi-armed bandit to avoid situations where a single timing out dialer wreaks havoc with all dialers.

### DIFF
--- a/chained/dialer.go
+++ b/chained/dialer.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


chained/dialer.go
```diff
@@ -188,16 +193,6 @@ func defaultDialOrigin(op *ops.Op, ctx context.Context, p *proxy, network, addr
...
-	var timeout time.Duration
-	if deadline, set := ctx.Deadline(); set {
-		conn.SetDeadline(deadline)
-		// Set timeout based on our given deadline, minus a 2 second fudge factor
-		timeout := time.Until(deadline) - 2*time.Second
-		if timeout < 0 {
-			log.Errorf("Not enough time left for server to dial upstream within %v, return errUpstream immediately", timeout)
-			return nil, errUpstream
-		}
-	}
```
myleshorton, line 200: This was just kind of wrong. If the context's deadline is exceeded, it's just done, so we should return. Otherwise, though, we don't want to mess around with the connection's read and write deadlines, which are separate from the context deadline, as the context deadline in net.DialContext land only governs the dial, not reads and writes, and presumably we want to adhere to that standard notion.

chained/dialer.go
```diff
@@ -256,9 +249,6 @@ func (p *proxy) buildCONNECTRequest(addr string, timeout time.Duration) (*http.R
 		Host: addr,
 	}
 	req.Host = addr
-	if timeout != 0 {
-		req.Header.Set(common.ProxyDialTimeoutHeader, fmt.Sprint(int(math.Ceil(timeout.Seconds()*1000))))
-	}
```
myleshorton, line 261: This is never used on the proxy, so the timeout passing here is unnecessary.

myleshorton:
The test error here is also on main and needs attention, but isn't related to this PR.

myleshorton:
Closing in favor of https://github.com/getlantern/flashlight/pull/1361

PR #1361, opened 2024-02-28
Only use a single dialer on each bandit dial

myleshorton:
This takes an alternative approach to simplifying the dialing logic, which is currently quite hampered by some proxies timing out.


bandit/bandit.go
```diff
@@ -56,13 +56,36 @@ func NewWithStats(dialers []Dialer, statsTracker stats.Tracker) (*BanditDialer,
...
+// as a means of seeding the bandit with initial data. This should
+// help weed out dialers/proxies censored users can't connect to at all.
+func parallelDial(dialers []Dialer, bandit *bandit.EpsilonGreedy) {
+	for index, d := range dialers {
+		// Note that the index of the dialer in our list is the index of the arm
+		// in the bandit and that the bandit is concurrent-safe.
+		go func(dialer Dialer, index int) {
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			_, _, err := dialer.DialContext(ctx, "tcp", "www.doesnotmatter.com")
```
hwh33: We should close the connection if the dial is successful.  Otherwise, we leak the file descriptor, right?

bandit/bandit.go
```diff
@@ -72,77 +95,66 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
+	// We don't give any special reward for a successful dial here and just rely on
+	// the normalized raw throughput to determine the reward. This is because the
+	// reward system takes into account how many tries there have been for a given
+	// "arm", so giving a reward here would be double-counting.
+
+	// Tell the dialer to update the bandit with it's throughput after 5 seconds.
+	dt := newDataTrackingConn(conn)
+	time.AfterFunc(secondsForSample*time.Second, func() {
+		speed := normalizeReceiveSpeed(dt.dataRecv)
+		//log.Debugf("Dialer %v received %v bytes in %v seconds, normalized speed: %v", dialer.Name(), dt.dataRecv, secondsForSample, speed)
```
hwh33, line 128: Commented out code

bandit/bandit.go
```diff
@@ -72,77 +95,66 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
+	log.Debugf("Dialer %v dialed in %v seconds", dialer.Name(), time.Since(start).Seconds())
+	// We don't give any special reward for a successful dial here and just rely on
+	// the normalized raw throughput to determine the reward. This is because the
+	// reward system takes into account how many tries there have been for a given
+	// "arm", so giving a reward here would be double-counting.
+
+	// Tell the dialer to update the bandit with it's throughput after 5 seconds.
+	dt := newDataTrackingConn(conn)
+	time.AfterFunc(secondsForSample*time.Second, func() {
+		speed := normalizeReceiveSpeed(dt.dataRecv)
```
hwh33: This is not concurrency safe (you're reading from a variable another routine may be writing to).  To be honest, I'm not clear on whether this is really a problem in this case.  It's not a big deal if you read outdated code.  If you read _during_ a write, will you get a completely bogus value?  I have no idea.  I don't know whether we use the ThreadSanitizer, but technically [any implementation of the runtime is allowed to halt execution of the program if it notices a race](https://go.dev/ref/mem#restrictions).

At the least, it would prevent us from testing this code with the race detector, or from using the ThreadSanitizer (I think we at least used to use this?).

bandit/bandit.go
hwh33: This might be useful to log honestly.

bandit/bandit.go
```diff
@@ -72,77 +95,66 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
+		//log.Debugf("Dialer %v received %v bytes in %v seconds, normalized speed: %v", dialer.Name(), dt.dataRecv, secondsForSample, speed)
+		o.bandit.Update(chosenArm, speed)
+	})
+	o.onSuccess(dialer)
+	return dt, err
+}
+
+func (o *BanditDialer) chooseDialerForDomain(dialers []Dialer, network, addr string) (Dialer, int) {
```
hwh33, line 135: What's the deal with this?  Is this for the ChatGPT domain shadowing thing?  I feel like this adds a lot of complexity when generally speaking the upstream domain should be irrelevant to whether a given proxy works for a user.  If we want to do domain-specific stuff, we should just do that separately IMO.

bandit/bandit.go
```diff
@@ -72,77 +95,66 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
-	if numDialers == 1 {
-		log.Debugf("Only one dialer, so returning the existing arm: %v", existingArm)
-		return existingArm
-	}
+	// We try to choose a different arm randomly.
 	for i := 0; i < 20; i++ {
-		newArm := eg.SelectArm(rand.Float64())
+		// This selects a new arm randomly with protobability 1.0
+		// of choosing exploration vs exploitation.
+		newArm := eg.SelectArm(1.0)
```
hwh33: Won't this always return the same arm...?
> SelectArm chooses an arm that exploits if the value is more than the epsilon
threshold, and explore if the value is less than epsilon

Am I reading that wrong?

hwh33: Seems like yes?  (`max` is deterministic)
```
func (b *EpsilonGreedy) SelectArm(probability float64) int {
	b.RLock()
	defer b.RUnlock()

	// Exploit
	if probability > b.Epsilon {
		return max(b.Rewards...)
	}

	// Explore
	return rand.Intn(len(b.Rewards))
}
```

bandit/bandit.go
```diff
@@ -163,16 +175,11 @@ func (o *BanditDialer) onFailure() {
 
 const secondsForSample = 6
 
-// 1 Mbps in bytes per second is the upper end of speeds we expect to see in such a
-// short time period.
+// A reasonable upper bound for the top expected bytes to receive in a short
+// window. Anything over this will be normalized to over 1.
 const topExpectedBytes = 125000
```
hwh33: I would call this `topExpectedBps` or `topExpectedBytesPerSecond`.

myleshorton:
Fixing this data race that's appearing in the tests...interesting

chained/dialer.go
```diff
@@ -188,16 +191,6 @@ func defaultDialOrigin(op *ops.Op, ctx context.Context, p *proxy, network, addr
...
 	}
 
-	var timeout time.Duration
-	if deadline, set := ctx.Deadline(); set {
-		conn.SetDeadline(deadline)
-		// Set timeout based on our given deadline, minus a 2 second fudge factor
-		timeout := time.Until(deadline) - 2*time.Second
-		if timeout < 0 {
-			log.Errorf("Not enough time left for server to dial upstream within %v, return errUpstream immediately", timeout)
-			return nil, errUpstream
```
hwh33, line 198: Ah, yeah, good to fix this!

hwh33: For those following along at home: this was the subject of a discussion on Slack yesterday.  We don't want to treat timeouts as upstream errors (where "upstream error" roughly translates to "not an issue with the proxy") as timeouts are often indicative of blocking (because the censor drops the connection in an on-path middle-box).

chained/dialer.go
```diff
@@ -180,6 +177,12 @@ func (p *proxy) MarkFailure() {
 // involves sending either a CONNECT request or a GET request to initiate a
 // persistent connection to the upstream proxy.
 func defaultDialOrigin(op *ops.Op, ctx context.Context, p *proxy, network, addr string) (net.Conn, error) {
+	select {
+	case <-ctx.Done():
+		log.Debug("Context done for dialing proxy address, returning error.")
+		return nil, ctx.Err()
+	default:
+	}
```
hwh33: This shouldn't be necessary.  Callers should be able to distinguish this with `errors.Is(err, context.DeadlineExceeded)`.  I guess this doesn't hurt though.

hwh33: If the goal is just to log here when that happens, IMO you would be better off (1) always logging the deadline, regardless of whether it has passed and (2) checking the error returned by `reportedDial` below to see if the dial timed out.  This would allow you to also see cases in which the dial was not given sufficient time to complete.

bandit/bandit.go
hwh33: Wait doesn't the domain matter?  What happens if doesnotmatter.com fails to respond?  Doesn't that result in an error here?

myleshorton: Great point on the file desc! The domain shouldn't matter because the dialer here is just dialing the proxy, but maybe that depends on the transport? I'll check that.

bandit/bandit.go
myleshorton: Great catch will fix!

myleshorton: Should be all set!!

chained/dialer.go
myleshorton: OK I'll just kill it for now...theoretically with these changes we shouldn't see those weird negative timeouts...

bandit/bandit.go
myleshorton: Ah right I wanted to go in the other direction (low probability of exploitation). Great catch!

bandit/bandit.go
myleshorton: No, it's that some of our dialers (aka broflake) are only available for some domains.

bandit/bandit.go
myleshorton: Yeah that was there before, and it is useful, but just chatty if you leave it on all the time.

bandit/bandit.go
myleshorton: Hmn...or maybe I should expose a dial server function or something...interesting

myleshorton:
OK I think this now addresses all your comment @hwh33? I ended up exposing dialing just to the proxy and attempted to simplify things a little bit, although I still think we could unify the dialers, proxyImpls, etc, but we can leave that for another day.

Thanks for the great review!!

chained/chained_test.go
```diff
@@ -164,11 +164,10 @@ func TestBadAddressToServer(t *testing.T) {
 		return nil, fmt.Errorf("fail intentionally")
 	}}
 	l := startServer(t)
-	req, err := p.buildCONNECTRequest("somebadaddressasdfdasfds.asdfasdf.dfads:532400", 5*time.Second+2*time.Millisecond+1*time.Nanosecond)
+	req, err := p.buildCONNECTRequest("somebadaddressasdfdasfds.asdfasdf.dfads:532400")
 	if err != nil {
 		t.Fatalf("Unable to build request: %s", err)
 	}
-	assert.Equal(t, "5003", req.Header.Get(common.ProxyDialTimeoutHeader))
```
myleshorton, line 171: Not used on the server, or anywhere

chained/dialer.go
```diff
@@ -176,10 +173,10 @@ func (p *proxy) MarkFailure() {
 	log.Tracef("Dialer %s consecutive failures: %d -> %d", p.Label(), newCF-1, newCF)
 }
 
-// defaultDialOrigin implements the method from serverConn. With standard proxies, this
+// dialOrigin implements the method from serverConn. With standard proxies, this
 // involves sending either a CONNECT request or a GET request to initiate a
 // persistent connection to the upstream proxy.
-func defaultDialOrigin(op *ops.Op, ctx context.Context, p *proxy, network, addr string) (net.Conn, error) {
+func dialOrigin(op *ops.Op, ctx context.Context, p *proxy, network, addr string) (net.Conn, error) {
```
myleshorton, line 179: This was always used everywhere, so just use it instead of having a field in proxy

bandit/bandit.go
hwh33: Could we handle that at a higher level?  Encoding that in the rewards system of the MAB dialer is pretty roundabout.

One possible way to handle this at a higher level would be maintaining one MAB dialer with Broflake and one without.  The client could then choose which to use based on the upstream domain.

bandit/bandit.go
```diff
@@ -72,83 +92,73 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
+	// We try to choose a different arm randomly.
 	for i := 0; i < 20; i++ {
-		newArm := eg.SelectArm(rand.Float64())
+		// This selects a new arm randomly, essentially forcing exploration by
+		// using a low probability vs a random probability.
+		newArm := eg.SelectArm(0.001)
 		if newArm != existingArm {
 			return newArm
 		}
 	}
```
hwh33: I'm confused about the purpose of this function to be honest.  It reduces to `rand.Intn(numDialers)`, but in a strange and verbose way.  Why not just rotate through the dialers until we find one we like?

Better yet, if we handle dialer preference at a higher level (as suggested above), we can do away with all of this stuff, right?

bandit/bandit.go
```diff
@@ -184,33 +189,36 @@ func (o *BanditDialer) Close() {
...
-	dataSent uint64
-	dataRecv uint64
+	dataRecv *atomic.Uint64
 }
 
 func (c *dataTrackingConn) Write(b []byte) (int, error) {
 	n, err := c.Conn.Write(b)
-	c.dataSent += uint64(n)
 	return n, err
 }
```
hwh33: You can just do away with this function now.

chained/multipath_test.go
```diff
@@ -0,0 +1,9 @@
+package chained
+
+import (
+	"testing"
+)
+
+func Test_groupByMultipathEndpoint(t *testing.T) {
+
+}
```
hwh33: ?

client/client_test.go
```diff
@@ -142,25 +141,24 @@ func TestServeHTTPTimeout(t *testing.T) {
 	req, err := http.NewRequest("CONNECT", "https://a.com:443", nil)
 	require.NoError(t, err)
 	resp, err := roundTrip(client, req)
-	_, ok := err.(*bandit.BanditError)
-	require.True(t, ok, "should return a BanditError")
+	//_, ok := err.(*bandit.BanditError)
+	//require.True(t, ok, "should return a BanditError but got %v", err)
```
hwh33: Should this just be removed?

hwh33 (commented):
The primary purpose of the PR (dialing only one "arm" on each dial) looks good, but the extra stuff with the domain-specific logic is not quite right IMO

myleshorton:
Hmnn ok the domain specific logic isn't new to this PR though - that was in the old balancer and the old bandit

client/client_test.go
myleshorton: oh yes will do!

chained/multipath_test.go
myleshorton: committed by mistake will fix!

hwh33 (approved):
> Hmnn ok the domain specific logic isn't new to this PR though

Oh, sorry, I see now.  You just refactored that into a new function.  Okay, yeah that's fine then.  I do stand by that being clunky with the MAB dialer though - I could see how the old balancer could at least apply the logic directly, though it does feel like the wrong layer.

myleshorton:
Thanks @hwh33! I agree with you on handling that at a higher level, btw, but the best design isn't immediately obvious to me, so I think we should handle it on another pass.

PR #1362, opened 2024-03-04
Remove iOS code

atavism:


jigar-f:
@atavism even after merging this, old ios client will still work correct?

atavism:
> @atavism even after merging this, old ios client will still work correct?

No.. do we need to support it? If we need to add back the ios code to flashlight for any reason, we can always reference this PR

jigar-f:
Even if we upload this build to production we still meed support for old clients, till most of user update app to latest

atavism:
> Even if we upload this build to production we still meed support for old clients, till most of user update app to latest

@jigar-f Oh, I see what you mean. Old versions of the app won't stop working (those builds will still be using previous versions of flashlight that included the iOS client code). So we should be fine merging this

jigar-f:
@atavism Thanks for letting me know, I will merge this shortly.

jigar-f:
@atavism Here are the changes-: https://github.com/getlantern/flashlight/pull/1362/commits/8bd9703242d23e89d1571503f42d4f81683cccb7 If that works for you.

bandit/bandit.go
```diff
@@ -34,6 +30,9 @@ type BanditDialer struct {
 	dialers      []Dialer
 	bandit       *bandit.EpsilonGreedy
 	statsTracker stats.Tracker
+	// lastDialCountryCode stores the last dialed country code.
+	// This is used to avoid unnecessary calls to the UI thread
+	lastDialCountryCode atomic.Value
```
jigar-f: Could you take a look at it here, Does this break any other code?
cc @atavism @reflog

bandit/bandit.go
```diff
@@ -160,6 +163,13 @@ func differentArm(existingArm, numDialers int) int {
 
 func (o *BanditDialer) onSuccess(dialer Dialer) {
 	countryCode, country, city := dialer.Location()
+	lastCountryCode, _ := o.lastDialCountryCode.Load().(string)
+	if lastCountryCode != "" && lastCountryCode == countryCode {
+		log.Debugf("Same country code as last returning")
```
reflog: let's remove this log. too noisy

reflog (approved):
👍🏻  (except for the log comment)

bandit/bandit.go
jigar-f: Thanks, Done.

PR #1374, opened 2024-04-19
add tls for algeneva

garmr-ulfr:


garmr-ulfr:
Test failing due to the updated version of lantern-algeneva not being merged into http-proxy yet.

chained/algeneva_impl.go
```diff
@@ -20,13 +24,34 @@ type algenevaImpl struct {
...
+		if block.Type != "CERTIFICATE" {
+			return nil, errors.New("expected certificate in PEM block")
+		}
+
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM([]byte(cert))
+		opts.TLSConfig = &tls.Config{
+			RootCAs:    certPool,
+			ServerName: pc.TLSServerNameIndicator,
+		}
```
hwh33: This would normally be the correct way to set this up.  However, the proxy cert is not signed for the hostname configured in `pc.TLSServerNameIndicator`.  That field is used to set a decoy value to throw off passive observers since the TLS server name indicator is transmitted in the clear.  I think you could just set this to the public IP of the proxy (which you can grab by parsing `addr`).

For reference, [we usually pin the proxy's certificate](https://github.com/getlantern/flashlight/blob/b36afb1d310a894232c26a31adefbc66eb419216/chained/tls_config.go#L74-L114).  We do this because Go doesn't offer a mechanism for both (1) advertising the SNI we want and (2) ignoring the mismatch between the SNI and the hostname in the proxy's certificate.  In the case of `algeneva`, it is not important that we advertise the bogus SNI.

hwh33 (approved):
Cool, looks good once the http-proxy-lantern reference is fixed.

PR #1376, opened 2024-04-26
Fix payment-v4 Test case

jigar-f:


pro/client/client_test.go
```diff
@@ -88,6 +89,9 @@ func TestPaymentMethodsV4(t *testing.T) {
 	user.UserID = res.User.ID
 	user.Token = res.User.Token
 
+	defer gock.Off()
+	mockGetRequest("/plans-v4", expectedPlansV4Response)
```
hwh33, line 93: Maybe you already realize this, but `UserCreate` above also hits staging and also needs to be mocked.  But that's the case for basically every test in this file... I guess we need to decide how much of this thread we want to pull.

jigar-f: Thanks @hwh33 for looking, But I forgot that we are working on #1371 PR that removed the proClient code and moves it to internalsdk, eventually we will write a test on internalsdk.

PR #1381, opened 2024-05-14
fix: add support to shadowsocks with TLS

WendelHime:


chained/shadowsocks_impl.go
```diff
@@ -75,11 +81,49 @@ func newShadowsocksImpl(name, addr string, pc *config.ProxyConfig, reportDialCor
...
+		}
+
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM([]byte(cert))
+		ip, _, _ := net.SplitHostPort(addr)
+		tlsConfig = &tls.Config{
+			RootCAs:               certPool,
+			ServerName:            ip,
+			InsecureSkipVerify:    true,
+			VerifyPeerCertificate: verifyPeerCert,
```
hwh33: So... I might have led you astray.  Thinking about this more, I think we might be able to just include the proxy certificate in `RootCAs`, as you have done, and skip the cert pinning / `VerifyPeerCertificate` stuff.  The proxy certificate is self-signed, so I think verification should pass if the proxy certificate is in the CA pool.

The reason we cannot do this normally is because we want to specify a bogus SNI (to throw off censors).  This bogus SNI causes normal verification to fail.  In this case, the TLS layer is wrapped in Shadowsocks, so the SNI will not be visible anyway.

Sorry for the mix-up!

hwh33: In short: I think we can do away with `VerifyPeerCertificate` and `InsecureSkipVerify` and keep everything else as-is.

hwh33 (commented):
Great work @WendelHime.  I actually misled you unfortunately - I left a comment explaining this.  Otherwise, this all looks awesome.

hwh33 (approved):
This looks great to me @WendelHime!  As discussed yesterday, we will wait until we've tested this before merging.

hwh33:
The CI failure looks like a flaky and unrelated test.  I think there's been some work to improve that test with mocks, but I'm not sure where that stands.  For now, I'm just re-running the test job.

PR #1382, opened 2024-05-28
Skip tests that aren't stable and remove confusing config tests

anacrolix:
Removes the snapshot config tests that Ox and I added for reasons that don't seem worthwhile anymore.

Skip tests that are flaky by default. They're still accessible if needed.


atavism (approved):

atavism:
LGTM!

PR #1384, opened 2024-06-03
Replacing lantern-shadowsocks implementation

WendelHime:
This PR requires https://github.com/getlantern/http-proxy/pull/619 to be merged so we can proceed with updating the mod references to http-proxy and use it on integration tests.

- [X] Updated and replaced references
- [x] E2E tests


hwh33 (approved):
LGTM.  Nice work @WendelHime!

In addition to testing this in combination with your changes, we should test:
1. An updated client against an old version of http-proxy.
2. An old client against an updated version of http-proxy.

This will tell us whether we need to version-gate this update.  I'm pretty sure both tests will succeed and we won't need to, but better safe than sorry!

hwh33:
Oh, actually there might be incompatibility in the config due to the new format of cipher strings ("chacha20-ietf-poly1305" to "AEAD_CHACHA20_POLY1305").  That's unfortunate =/

hwh33:
We can fix the incompatibility by having the client interpret "chacha20-ietf-poly1305" as "AEAD_CHACHA20_POLY1305".  I think we should do that so that we don't have to segment clients.  What do you think @WendelHime?

hwh33:
CI is failing due to a flaky test.  @anacrolix recently added logic to skip flaky tests in https://github.com/getlantern/flashlight/pull/1382.  It looks like your branch doesn't contain those changes.  If you merge in main, that should fix the CI failure.

WendelHime:
> LGTM. Nice work @WendelHime!
> 
> In addition to testing this in combination with your changes, we should test:
> 
> 1. An updated client against an old version of http-proxy.
> 2. An old client against an updated version of http-proxy.
> 
> This will tell us whether we need to version-gate this update. I'm pretty sure both tests will succeed and we won't need to, but better safe than sorry!

- Using the new client against the old http-proxy worked
- Using the old client against the http-proxy:feat-1380 worked too!

> We can fix the incompatibility by having the client interpret "chacha20-ietf-poly1305" as "AEAD_CHACHA20_POLY1305". I think we should do that so that we don't have to segment clients. What do you think @WendelHime?

Looking at the outline-sdk library they made it compatible [here](https://github.com/Jigsaw-Code/outline-sdk/blob/main/transport/shadowsocks/cipher.go#L70-L71), so I guess it won't be required to change it

hwh33:
> - Using the new client against the old http-proxy worked
> - Using the old client against the http-proxy:feat-1380 worked too!

Nice, thanks for testing both of those cases!

>> We can fix the incompatibility by having the client interpret "chacha20-ietf-poly1305" as "AEAD_CHACHA20_POLY1305". I think we should do that so that we don't have to segment clients. What do you think @WendelHime?
> 
> Looking at the outline-sdk library they made it compatible [here](https://github.com/Jigsaw-Code/outline-sdk/blob/main/transport/shadowsocks/cipher.go#L70-L71), so I guess it won't be required to change it

Oh, great!  Good catch.

PR #1386, opened 2024-06-05
Enable chat for small percentage of Android users in Iran

atavism:
Resolves https://github.com/getlantern/engineering/issues/1385

To make it easier to test this, chat is also enabled now by default when running a dev build


jigar-f (approved):

atavism:
I double checked chat works and the tab shows up if the feature is enabled [here](https://github.com/getlantern/lantern-client/blob/main/internalsdk/android.go#L592). That should work for existing clients (7.8.4 and up) so we should only have to make these config changes.

jigar-f:
Thank you for making these changes @atavism.

atavism:
Hey @hwh33, do these changes look good to you too?

PR #1372, opened 2024-04-15
Add callbacks for initialization states

atavism:

PR #1387, opened 2024-06-06
Add callbacks for initialization states

atavism:
New PR for https://github.com/getlantern/flashlight/pull/1372

### DIFF
--- a/bandit/bandit.go
+++ b/bandit/bandit.go
@@ -27,19 +27,36 @@ const (
 
 // BanditDialer is responsible for continually choosing the optimized dialer.
 type BanditDialer struct {
+	opts         Options
 	dialers      []Dialer
 	bandit       *bandit.EpsilonGreedy
 	statsTracker stats.Tracker
 }
 
-// New creates a new bandit given the available dialers.
-func New(dialers []Dialer) (*BanditDialer, error) {
-	return NewWithStats(dialers, stats.NewNoop())
+// Options are the options used to create a new bandit
+type Options struct {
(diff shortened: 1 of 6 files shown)


bandit/bandit.go
```diff
@@ -27,19 +27,36 @@ const (
 
 // BanditDialer is responsible for continually choosing the optimized dialer.
 type BanditDialer struct {
+	opts         Options
```
hwh33: We pull `opts.Dialers` out into `BanditDialer.dialers` and `opts.StatsTracker` out into `BanditDialer.statsTracker`.  Then we add `opts` directly onto `BanditDialer`, creating redundant fields.  Why don't we just also pull `opts.OnError` and `opts.OnSuccess` out directly onto `BanditDialer`?  This would also allow us to do nil checks in `New` (and create no-op functions for `BanditDialer.OnSuccess/Error`), rather than needing to do the nil checks at every call site.

hwh33 (approved):
Just one comment about a small code organization thing, but otherwise looks good to me.

bandit/bandit.go
atavism: Thanks, @hwh33! I moved opts.(onError/onSuccess) directly onto BanditDialer and the nil checks to New. How do the changes look to you now?

hwh33 (approved):
Looks great!

bandit/bandit.go
hwh33: Looks great!  Thanks!

atavism:
@hwh33 Thanks for the review!

PR #1389, opened 2024-06-11
Adding water dialer implementation

WendelHime:


chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestNewWaterImpl(t *testing.T) {
+	type params struct {
+		raddr          string
+		pc             *config.ProxyConfig
+		reportDialCore reportDialCoreFn
+	}
+
+	f, err := os.ReadFile("../wasm/reverse.go.wasm")
+	require.Nil(t, err)
```
garmr-ulfr: When checking if there's an error in tests you should use `Error`/`NoError` instead of `Nil`/`NotNil`.

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+					PluggableTransportSettings: map[string]string{
+						"water_wasm": string(f),
+					},
+				},
+				reportDialCore: func(op *ops.Op, dialCore func() (net.Conn, error)) (net.Conn, error) {
+					return nil, nil
+				},
+			},
+			assert: func(t *testing.T, actual *waterImpl, err error) {
+				assert.Nil(t, err)
```
garmr-ulfr: Same as above. This could also be `require` instead of `assert`. If there's an error then it's assumed that `actual` would be nil or invalid and shouldn't be used. That being said, `assert` is fine if you prefer it!

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+				assert.NotNil(t, actual)
```
garmr-ulfr: You should use `require.NotNil` here. If it is nil then the next line will panic.

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+		givenParams params
+		assert      func(t *testing.T, actual *waterImpl, err error)
+	}{
+		{
+			name: "create new waterImpl with success",
+			givenParams: params{
+				raddr: "127.0.0.1",
+				pc: &config.ProxyConfig{
+					PluggableTransportSettings: map[string]string{
+						"water_wasm": string(f),
```
garmr-ulfr: Since what you're testing doesn't rely on the file or it's contents, I would just set "water_wasm" to some random string instead of reading from a file. You're way works too though!

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+					},
+				},
+				reportDialCore: func(op *ops.Op, dialCore func() (net.Conn, error)) (net.Conn, error) {
+					return nil, nil
+				},
+			},
+			assert: func(t *testing.T, actual *waterImpl, err error) {
+				assert.Nil(t, err)
+				assert.NotNil(t, actual)
+				assert.NotNil(t, actual.config)
```
garmr-ulfr: This will pass as long as `actual.config` is not nil, even if it's not correct. You should create a `wantConfig` to compare against using `assert.Equal`. 
```go
wasm := "test wasm"
var tests = []struct {
  name        string
  givenParams params
  wantConfig *water.Config
  assert      func(t *testing.T, actual *waterImpl, err error)
}{
    {
      ...
        pc: &config.ProxyConfig{
          PluggableTransportSettings: map[string]string{
	    "water_wasm": wasm,
          },
        },
      ...
    },
    wantConfig: &water.Config{
      TransportModuleBin: []byte(wasm),
    },
    assert: func(t *testing.T, wantConfig *water.Config, actual *waterImpl, err error) {
      require.NoError(t, err)
      require.NotNil(t, actual)

      assert.Equal(t, wantConfig, actual.config)
      assert.NotNil(t, actual.reportDialCore)
    },
  },
}
...
    tt.assert(t, tt.wantConfig, waterImpl, err)
...
}
```

This will also print `wantConfig` and `actual.config` if it fails, making it easier to debug.

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			waterImpl, err := newWaterImpl(tt.givenParams.raddr, tt.givenParams.pc, tt.givenParams.reportDialCore)
+			tt.assert(t, waterImpl, err)
+		})
+	}
+}
+
+func TestWaterDialServer(t *testing.T) {
+	f, err := os.ReadFile("../wasm/reverse.go.wasm")
+	require.Nil(t, err)
```
garmr-ulfr: Same as above regarding `Error`/`NoError` functions.

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+				assert.NotNil(t, err)
+				assert.ErrorContains(t, err, "transport endpoint is not connected")
+				assert.Nil(t, actual)
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			waterImpl, err := newWaterImpl("127.0.0.1:8080", pc, tt.givenReportDialCore)
+			require.Nil(t, err)
```
garmr-ulfr: Same as above regarding `Error`/`NoError` functions.

chained/water_impl_test.go
```diff
@@ -0,0 +1,100 @@
...
+			givenOp:  testOp,
+			givenCtx: context.Background(),
+			givenReportDialCore: func(op *ops.Op, dialCore func() (net.Conn, error)) (net.Conn, error) {
+				assert.Equal(t, testOp, op)
+				assert.NotNil(t, dialCore)
+				return dialCore()
+			},
+			assert: func(t *testing.T, actual net.Conn, err error) {
+				assert.NotNil(t, err)
+				assert.ErrorContains(t, err, "transport endpoint is not connected")
```
garmr-ulfr: `ErrorContains` also checks for nil so you don't need to explicitly check for it.

chained/water_impl.go
```diff
@@ -0,0 +1,41 @@
...
+			TransportModuleBin: wasm,
+		},
+		reportDialCore: reportDialCore,
+	}, nil
+}
+
+func (d *waterImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
+	dialer, err := water.NewDialerWithContext(ctx, d.config)
+	if err != nil {
+		return nil, errors.New("failed to create dialer: %v", err)
```
garmr-ulfr: It would be better to wrap the error instead of creating a new one. `fmt.Errorf("failed to create dialer: %w", err)`

garmr-ulfr (commented):
Looks good! Just a few things though.

chained/water_impl_test.go
WendelHime: Thanks for the suggestions! I've just implemented this a little bit different (defined `wantConfig` before tests  because I didn't want to pass as an assert argument since I'm just receiving the parameters returned from the function in scope).

chained/water_impl_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestNewWaterImpl(t *testing.T) {
+	type params struct {
+		raddr          string
+		pc             *config.ProxyConfig
+		reportDialCore reportDialCoreFn
+	}
+
+	randomWASM := make([]byte, 1024)
+	_, err := crand.Read(randomWASM)
+	require.Nil(t, err, "Failed to generate random WASM content")
```
garmr-ulfr: Missed one here. Should be `NoError`.

garmr-ulfr (approved):
LGTM

chained/common.go
```diff
@@ -78,6 +78,14 @@ func _settingBool(settings map[string]string, name string) bool {
 	return val
 }
 
+func _settingBytes(settings map[string]string, name string) []byte {
+	_val := _setting(settings, name)
+	if _val == "" {
+		return nil
+	}
+	return []byte(_val)
+}
```
hwh33: We won't be able to pass raw bytes in this config.  The config is sent and parsed a YAML - a text-based format.  Including raw UTF-encoded bytes will potentially (likely) screw up the encoding of the YAML document.

Similar to the flag passed to [http-proxy](https://github.com/getlantern/http-proxy/pull/622/files#diff-dddc5dd1f92ee518b251cc43ab54489f6e4a3132e6e857baecabb3d271fe8ab2R190), this will need to be encoded in a text-friendly format.  Base64 would work well here too.

hwh33: Oh, I see that you're treating the WASM as a base64-encoded string in water_impl.go.  So I guess you figured this out already 🙂 

I think this is dead code then, right?  Looks like it's no longer used, so we should just get rid of it.

chained/water_impl.go
```diff
@@ -0,0 +1,47 @@
...
+func newWaterImpl(addr string, pc *config.ProxyConfig, reportDialCore reportDialCoreFn) (*waterImpl, error) {
+	b64WASM := ptSetting(pc, "water_wasm")
+	wasm, err := base64.StdEncoding.DecodeString(b64WASM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode water wasm: %w", err)
+	}
+
+	return &waterImpl{
+		raddr: addr,
+		config: &water.Config{
```
hwh33: I think we want to provide a logger (using `water.Config.OverrideLogger`).  This should use `golog`, like the rest of flashlight's logs.

chained/water_impl_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+func TestWaterDialServer(t *testing.T) {
+	f, err := os.ReadFile("../wasm/reverse.go.wasm")
+	require.NoError(t, err)
+	b64WASM := base64.StdEncoding.EncodeToString(f)
+
+	pc := &config.ProxyConfig{PluggableTransportSettings: map[string]string{"water_wasm": b64WASM}}
+	testOp := ops.Begin("test")
```
hwh33, line 82: nit: This should have a `defer testOp.End()`

go.mod
```diff
@@ -20,6 +20,8 @@ replace github.com/eycorsican/go-tun2socks => github.com/getlantern/go-tun2socks
 // on this unmerged PR against gomobile - https://github.com/golang/mobile/pull/58.
 replace golang.org/x/mobile => github.com/oxtoacart/mobile v0.0.0-20220116191336-0bdf708b6d0f
 
+replace github.com/tetratelabs/wazero => github.com/refraction-networking/wazero v1.7.1-w
```
hwh33, line 23: I'm guessing this was done automatically when you ran `go get` for `water`.  Is that right?  Just wanted to make sure this is intentional.

hwh33: It's not typical, but it could be due to some way water pulls in this `wazero` package (I haven't looked).

wasm/plain.go.wasm
hwh33, line 1: Could you put these in `chained/testdata`?  (That's the typical location for test files in Go)

hwh33 (commented):
Just a few minor comments.  Nice work!

chained/water_impl.go
WendelHime: hmmmm in this case golog isn't accepted by OverrideLogger because it expects a `slog.Logger` and the `golog` doesn't implement that, I'll try to do this but Idk if it'll work (probably requires `TRACE` env var set to true):
```go
	return &waterImpl{
		raddr: addr,
		config: &water.Config{
			TransportModuleBin: wasm,
			OverrideLogger:     slog.New(slog.NewTextHandler(log.TraceOut(), nil)),
		},
		reportDialCore: reportDialCore,
	}, nil

```

go.mod
WendelHime: No I had to do it manually, `wazero` has a different implementation from `tetratelabs/wazero` that contain different/custom functions. If we use `tetratelabs/wazero`, `water` will try to access functions that doesn't exist and panic (and their version has the functions they're expecting). Every repository that refers to the WATER package will need this replacement.

chained/water_impl.go
hwh33: Yeah, you'll need to wrap `golog` to get it to fit.  Sorry if that was unclear.  Looks like you will need to provide a custom [slog.Handler](https://pkg.go.dev/log/slog#Handler).  Admittedly, that looks like more work than I originally anticipated.  I think it still looks reasonable though?  Could you give it a go and let me know if you run into any major issues?  I don't think we want to just have logs dumped to stdout or a framework outside of our normal logging framework.

go.mod
hwh33: I see!  Yeah, I see the same `replace` directive in the `water` library itself.  It's a shame that `go get` doesn't at least call this out when you pull in a new dependency (I tried it myself just to see).

I raised this issue with the WATER team [here on Slack](https://wdynhnkxvsdx.slack.com/archives/C06GJBFP8C8/p1719361086938969).  It's not a huge deal, but it seems like an easily resolvable pain point.  It doesn't look like `water` has any indirect dependencies on `wazero` so I believe they should be able to use their fork directly and do away with the `replace` statement.

Anyway, looks like we don't have a choice at the moment, so we can just leave it as is for now!  Thanks for explaining the situation.

chained/water_impl.go
WendelHime: I've added a small logger implementation following the `slog.Handler`, it doesn't implement all features like `WithGroup` but it should be enough for now. We should be able to see it working when we deploy the reverse test proxy with a E2E test.

WendelHime: I'll add some tests later after adding suggestions from http-proxy and lantern

chained/logger.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/getlantern/golog"
+)
+
+// slogHandler is a Handler that implements the slog.Handler interface
+// and writes log records to a golog.Logger.
+type slogHandler struct {
```
hwh33, line 14: Awesome, this is really well done

chained/water_impl.go
hwh33: Thanks for digging into this!

hwh33 (approved):
Looks great!  Nice work @WendelHime

chained/logger.go
```diff
@@ -12,15 +12,16 @@ import (
...
+	transport string
+	minLevel  slog.Level
+	opts      slog.HandlerOptions
+	attrs     string
+	groups    []string
 }
 
-func newLogHandler(logger golog.Logger) *slogHandler {
-	return &slogHandler{logger: logger}
+func newLogHandler(logger golog.Logger, transport string) *slogHandler {
```
hwh33: With this addition, this becomes a logger specific to use with Water (or at least transports).  Could you rename it as such?  Alternatively, I think you could keep it general by adding the transport as an attribute - though I like how you've done it already.  Or you could add support for an optional prefix instead of being specifically a transport and required.

chained/water_impl.go
```diff
@@ -0,0 +1,57 @@
...
+	return &waterImpl{
+		raddr:          addr,
+		dialer:         dialer,
+		reportDialCore: reportDialCore,
+	}, nil
+}
+
+func (d *waterImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
+	return d.reportDialCore(op, func() (net.Conn, error) {
+		conn, err := d.dialer.DialContext(context.Background(), "tcp", d.raddr)
```
hwh33, line 55: We do actually want to respect the dial context.  The issue is a bug in `water`.  I pinged the WATER team on Slack about it [here](https://wdynhnkxvsdx.slack.com/archives/C06GJBFP8C8/p1721341403567499?thread_ts=1720463188.453499&cid=C06GJBFP8C8).  Essentially the library hanging onto the dial context beyond its scope (just the dial).  If you cancel this context, all connections dialed with the context will be closed.  This should not happen - only dials in progress should be affected.

Here's a little example program to demonstrate, with a comparison to a simple TCP dialer/listener pair from the net package:

https://github.com/hwh33/water-context-bug

We can just mark this as a TODO for now though.  Could you add a TODO comment here explaining the bug and the reason for ignoring the dial context?  Then we can merge this PR and continue to make progress with integration.  When the WATER team fixes the issue, we can circle back and use `ctx` for the dial here.

hwh33 (approved):
Two minor changes requested, but otherwise LGTM.  Awesome work on this @WendelHime!!

chained/logger.go
WendelHime: I'll add support for a prefix!

chained/water_impl.go
hwh33: Comment looks great!

chained/logger.go
```diff
@@ -12,16 +12,16 @@ import (
...
+	minLevel slog.Level
+	opts     slog.HandlerOptions
+	attrs    string
+	groups   []string
 }
 
-func newLogHandler(logger golog.Logger, transport string) *slogHandler {
-	return &slogHandler{logger: logger, transport: transport}
+func newLogHandler(logger golog.Logger, prefix string) *slogHandler {
+	return &slogHandler{logger: logger, prefix: prefix}
```
hwh33, line 24: Nice!

hwh33 (approved):

PR #1391, opened 2024-06-19
updated stats when previousStats is nil.

jigar-f:


bandit/bandit.go
```diff
@@ -163,7 +163,7 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
 
 	countryCode, country, city := dialer.Location()
 	previousStats := o.statsTracker.Latest()
-	if previousStats.CountryCode != "" && previousStats.CountryCode != countryCode {
+	if previousStats.CountryCode == "" || previousStats.CountryCode != countryCode {
```
jigar-f, line 166: @reflog 

So the problem here is we were checking  `previousStats.CountryCode != ""` but using old code previousStats.CountryCode it always remains null because we were setting o.statsTracker.SetActiveProxyLocation. only when `previousStats.CountryCode != ""`  and without setting `o.statsTracker.SetActiveProxyLocation` previous state is always going to be null 


So basically with the old code, we were getting previousStats always null

jigar-f: So new code checks if `previousStats.CountryCode == ""` is null then set location so we on next call we will get the previous state not null and it will check for `previousStats.CountryCode != countryCode``

reflog (approved):

PR #1395, opened 2024-07-16
group client config requests

garmr-ulfr:
This groups the proxy, geolocation, and user config requests to reduce start up time. The config request service utilizes the `callRandomly` code from `bypass` that adds jitter between requests to prevent [the thundering herd problem](https://encore.dev/blog/thundering-herd-problem), as well as avoid creating a pattern that censors could detect.


garmr-ulfr:
There are still a couple of things that need to be done, such as update the original config fetcher to not fetch proxy configs, have flashlight start the new config and bypass services, add tests, etc.

apipb/legacy.go
```diff
@@ -0,0 +1,218 @@
+// Copied from lantern-cloud: e57d588aa976ca9d1a8531c92972e20a7d1640f9
```
hwh33, line 1: We talked about this in a 1:1, but I just want to double-check: this whole file is only needed temporarily, right?  Once we're using `apipb.ProxyConnectConfig` everywhere, we'll be able to do away with this?

config/config_test.go
```diff
@@ -106,105 +103,108 @@ func TestObfuscated(t *testing.T) {
...
-		fi, err = os.Stat(file)
-
-		assert.NotNil(t, fi)
-		assert.Nil(t, err, "Got error: %v", err)
-
-		assert.True(t, fi.ModTime().After(mtime))
-	})
-}
+/*
+	commented out for now because this is for the old proxy config stuff
```
hwh33: Will these eventually be deleted?  Replaced?

config/initializer.go
```diff
@@ -19,69 +17,54 @@ const packageLogPrefix = "flashlight.config"
...
-	// DefaultProxyConfigPollInterval determines how frequently to fetch proxies.yaml
-	DefaultProxyConfigPollInterval = 1 * time.Minute
-
-	// ForceProxyConfigPollInterval overrides how frequently to fetch proxies.yaml if set (does not honor values from global.yaml)
-	ForceProxyConfigPollInterval = 0 * time.Second
+	// // DefaultProxyConfigPollInterval determines how frequently to fetch proxies.yaml
+	// DefaultProxyConfigPollInterval = 1 * time.Minute
+	//
+	// // ForceProxyConfigPollInterval overrides how frequently to fetch proxies.yaml if set (does not honor values from global.yaml)
+	// ForceProxyConfigPollInterval = 0 * time.Second
```
hwh33: What's the plan with these?

config/initializer.go
```diff
@@ -90,9 +73,9 @@ func InitWithURLs(
 			if globalConfig.GlobalConfigPollInterval > 0 {
 				globalConfigPollInterval = globalConfig.GlobalConfigPollInterval
 			}
-			if ForceProxyConfigPollInterval == 0 && globalConfig.ProxyConfigPollInterval > 0 {
-				proxyConfigPollInterval = globalConfig.ProxyConfigPollInterval
-			}
+			// if ForceProxyConfigPollInterval == 0 && globalConfig.ProxyConfigPollInterval > 0 {
+			// 	proxyConfigPollInterval = globalConfig.ProxyConfigPollInterval
+			// }
```
hwh33: We should never need this code again, right?  Why not just delete it?

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
+// Description: This file contains the implementation of the new consolidated proxy/geolocation config.
```
hwh33: Could you use the conventional `// Package proxyconfig does a thing` syntax?

Unless I'm missing something!

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/getlantern/flashlight/v7/apipb"
+)
+
+const (
+	defaultConfigSaveDir  = ""
+	defaultConfigFilename = "proxies.conf"
+)
+
+// alias for better readability
+type ProxyConfig = apipb.ConfigResponse
```
hwh33: There's more to the config response than the proxy config though - for example, the pro token and eventually the entire global config.

Same goes for other names here: "proxies.conf", the package name, etc.

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+var (
+	_config = &config{
+		config: eventual.NewValue(),
+	}
+
+	log = golog.LoggerFor("proxyconfig")
+)
+
+type config struct {
+	config eventual.Value
```
hwh33: Could you use a comment to indicate the type held by this value?

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+// GetConfig implements services.ConfigHandler
+func (c *config) GetConfig() *ProxyConfig {
+	conf, _ := c.config.Get(eventual.DontWait)
+	return conf.(*ProxyConfig)
+}
+
+// SetConfig implements services.ConfigHandler
+func (c *config) SetConfig(new *ProxyConfig) {
+	old := c.GetConfig()
+	c.config.Set(new)
```
hwh33: What about parts of the config response which were not updated by the back-end?  I think you need some merge logic here.

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := saveConfig(c.filePath, new, c.obfuscate); err != nil {
+		log.Errorf("Failed to save client config: %w", err)
+	}
+
+	notifyListeners(old, new)
+}
+
+// GetConfig returns the current client config. An error is returned if the config is not available
+// within the given timeout.
+func GetConfig(timeout time.Duration) (*ProxyConfig, error) {
```
hwh33: This would be more flexible and idiomatic with a context (e.g. `GetConfig(ctx context.Context)`).  Should be easy enough since `eventual` uses contexts anyway.

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	conf, err := _config.config.Get(ctx)
```
hwh33: You need a `nil` check here.  This is true whether or not you take my suggestion above to take a context instead of a timeout.  As it stands, a user could provider a timeout of 0 and possibly end up with a nil `conf` and subsequent panic.

hwh33: To fully obscure `eventual` from users of this package, I would personally treat the `nil` case as an error, like:
```
conf, err := _config.config.Get(ctx)
if err != nil {
    return nil, fmt.Errorf("failed to get config: %w", err")
}
if conf == nil {
    // This can only be due to a combination of unset config and expired context.
    return nil, fmt.Errorf("config not yet set: %w", ctx.Error())
}
```

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+// GetProxyConfigs returns the list of proxy configs from the current client config. An error is
+// returned if the config is not available within the given timeout.
+func GetProxyConfigs(timeout time.Duration) ([]*apipb.ProxyConnectConfig, error) {
+	config, err := GetConfig(timeout)
+	if err != nil || config == nil {
+		return nil, err
+	}
+
+	return config.GetProxy().GetProxies(), nil
+}
```
hwh33: What do these functions accomplish?  With:
```
token, err := GetProToken(timeout)
if err != nil {
  // handle
}
// use token
```

Without:
```
cfg, err := GetConfig(timeout)
if err != nil {
  // handle err 
}
// use cfg.GetProToken()
```

Removing them will reduce the API surface area (IMO a good thing all else equal) and reduce the need for boilerplate as `apipb.ConfigResponse` is expanded.

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+		in = rot13.NewReader(infile)
+	}
+
+	bytes, err := io.ReadAll(in)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read config from %v: %w", filePath, err)
+	}
+
+	if len(bytes) == 0 {
+		return nil, nil // file is empty
```
hwh33: It's unusual for functions in Go to return an unusable value and no error.  I see where you're using it and it looks fine.  Could you just note in the documentation above this function that this is a possibility and the conditions by which it might occur?

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+
+	_config.listeners = append(_config.listeners, fn)
+	_config.mu.Unlock()
+}
+
+func notifyListeners(old, new *ProxyConfig) {
+	_config.mu.Lock()
+	listeners := _config.listeners
+	_config.mu.Unlock()
+	// TODO: should we clone the configs before passing them to the listeners?
```
hwh33: Good question.  I say let's do it, for good measure.  Since the config is a protocol buffer message, it's conveniently quite easy with [proto.Clone](https://pkg.go.dev/google.golang.org/protobuf/proto#Clone).

config/proxy/proxy.go
```diff
@@ -0,0 +1,197 @@
...
+	for _, fn := range listeners {
+		// don't block the config service
+		go fn(old, new)
```
hwh33: Good call!

flashlight.go
```diff
@@ -437,46 +437,137 @@ func (f *Flashlight) Run(httpProxyAddr, socksProxyAddr string,
...
-		stopConfigFetch()
 		stopMonitor()
 		stopBypass()
+		stopGlobalConfigFetch()
+		stopConfigService()
+	}, nil
+}
+
+func (f *Flashlight) startConfigService() (services.StopFn, error) {
+	obfuscate, _ := f.flagsAsMap["readableconfig"].(bool)
```
hwh33: Wouldn't the zero value be `nil` and therefore not castable to `bool`?

flashlight.go
```diff
@@ -437,46 +437,137 @@ func (f *Flashlight) Run(httpProxyAddr, socksProxyAddr string,
...
 	}
+
+	fn := func(old, new *proxyconfig.ProxyConfig) {
+		var country string
+		if old != nil {
+			country = old.GetCountry()
+		}
+
+		// update the country if it has changed
+		if nc := new.GetCountry(); nc != country {
```
hwh33: Looks like `new` can be `nil` (for example, in the call below on line 503).  This would result in a panic here.

hwh33: It'd be nice actually if `proxyconfig.OnConfigChange` could guarantee that `new` will never be nil.  You'd still have to fix the direct call to `fn` on line 503, but it'd make life easier for every caller of `proxyconfig.OnConfigChange`.

flashlight.go
```diff
@@ -437,46 +437,137 @@ func (f *Flashlight) Run(httpProxyAddr, socksProxyAddr string,
...
+	}
+
+	setConfigFlagOpts(configOpts, f.flagsAsMap)
+	return services.StartConfigService(handler, configOpts)
+}
+
+// setConfigFlagOpts sets the OriginURL, Sticky, and Obfuscate config options based on the input flags.
+func setConfigFlagOpts(opts *services.ConfigOptions, flags map[string]interface{}) {
+	toBool := func(v interface{}) bool {
+		b, _ := v.(bool)
```
hwh33: Similar to above, I don't believe this is safe.  When `v` is nil, I believe this would panic.  Happy to be proven wrong!  Maybe we have some logic which sets a default value for every flag?

hwh33: > Maybe we have some logic which sets a default value for every flag?

In that case though, I would prefer to have nil checks here.  Relying here on the behavior of our flag parsing logic would be fragile.

geolookup/geolookup.go
```diff
@@ -2,261 +2,80 @@ package geolookup
...
+	proxyconfig.OnConfigChange(func(old, new *proxyconfig.ProxyConfig) {
+		oldCCountry, _ := _country.Get(eventual.DontWait)
+		oldIP, _ := _ip.Get(eventual.DontWait)
+
+		_country.Set(new.Country)
+		_ip.Set(new.Ip)
+
+		// if the country or IP has changed, notify watchers
+		if oldCCountry != new.Country || oldIP != new.Ip {
+			for _, ch := range watchers {
```
hwh33, line 31: You need to lock `mx` before accessing `watchers`, right?

services/service.go
```diff
@@ -0,0 +1,49 @@
+package services
```
hwh33, line 5: Could you add a package doc comment?

services/bypass.go
```diff
@@ -0,0 +1,283 @@
+package services
+
+// bypass periodically sends traffic to the bypass blocking detection server. The server uses the ratio
+// between domain fronted and proxied traffic to determine if proxies are blocked. The client randomizes
+// the intervals between calls to the server and also randomizes the length of requests.
```
hwh33: This comment should be after the import statements I think.

services/service.go
```diff
@@ -0,0 +1,49 @@
...
+	jitter time.Duration,
+	done <-chan struct{},
+	logger golog.Logger,
+) {
+	intervalMin := int64(interval - jitter)
+	intervalMax := int64(interval + jitter)
+
+	// calculate sleep time
+	sleep := func(extraDelay time.Duration) <-chan time.Time {
+		delay := mrand.Int63n(intervalMax) + intervalMin
```
hwh33: This would need to be:
```
delay := mrand.Int63n(intervalMax-intervalMin) + intervalMin
```

Alternatively:
```
delay := mrand.Int63n(2*jitter) + interval - jitter
```

services/service.go
```diff
@@ -0,0 +1,49 @@
+package services
+
+import (
+	"time"
+
+	mrand "math/rand"
```
hwh33: Let's use the new version of this package (just learned about it myself!):
https://pkg.go.dev/math/rand/v2

services/service.go
```diff
@@ -0,0 +1,49 @@
...
+	for {
+		select {
+		case <-done:
+			return
+		case <-sleep(extraDelay):
+			start := time.Now()
+			extraSeconds = fn()
+			elapsed := time.Since(start)
+			// aren't we supposed to subtract elapsed from extraDelay instead of adding?
+			extraDelay = time.Duration(extraSeconds)*time.Second + elapsed
```
hwh33: Heh, yeah looks like that was the original intention:
https://github.com/getlantern/flashlight/pull/1232#discussion_r911714670

But then what happens if this causes `extraDelay` to hit zero (or go negative, which we should of course adjust to zero)?  I'm not sure what's correct in that case to be honest.  Given our long sleep times, that case is not terribly likely.  So I guess for now, let's just subtract `elapsed` and ensure `extraDelay` never goes below zero.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+type ConfigOptions struct {
+	// URL to use for fetching this config.
+	OriginURL string
+
+	// UserConfig contains data for communicating the user details to upstream
+	// servers in HTTP headers, such as the pro token and other options.
+	UserConfig common.UserConfig
+
+	// Sticky specifies whether or not to only use the local config and not
+	// update it with remote data.
```
hwh33: Not sure if this comment was ported from somewhere else, but I find the near double negative confusing.  What about:

"Sticky specifies whether to stick with the local config. If true, remote config will be ignored."

hwh33: Actually, this guarantee is not possible with your current API.  You don't have any control in this package over the first result of `ConfigHandler.GetConfig` (since it's an interface with no strict requirements on the source of the first call).

I think this needs more thought.  I'm not sure what the solution is to be honest.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	Sticky bool
+
+	// RoundTripper provides the http.RoundTripper the fetcher should use, which allows us to
+	// dictate whether the fetcher will use dual fetching (from fronted and chained URLs) or not.
+	RoundTripper http.RoundTripper
+
+	// PollInterval specifies how frequently to poll for new config.
+	PollInterval time.Duration
+	// PollJitter specifies the max amount of jitter to add to the poll interval.
+	PollJitter time.Duration
```
hwh33: I don't think this needs to be configurable.  Let's just let `services` handle this.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	fn := func() int64 {
+		sleep, _ := _configService.fetchConfig()
+		return sleep
+	}
+	go callRandomly(fn, opts.PollInterval, opts.PollJitter, _configService.done, _configService.logger)
+
+	return _configService.Stop, nil
+}
+
+func (cs *configService) updateClientInfo(conf *apipb.ConfigResponse) {
```
hwh33: This is only called once and is quite minimal.  I would just move this code to the call site.  This is particularly helpful for analyzing the concurrency concerns with `configService`.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+
+	if opts.PollInterval <= 0 {
+		opts.PollInterval = defaultConfigPollInterval
+	}
+
+	if opts.PollJitter <= 0 {
+		opts.PollJitter = defaultConfigPollJitter
+	}
+
+	logger := golog.LoggerFor("configservice")
```
hwh33: We initialize loggers on the package level, not the type level.  Could you move this to package scope and re-name the prefix?

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+
+	u, err := url.Parse(opts.OriginURL)
+	if err != nil {
+		logger.Fatalf("Unable to parse chained cloud config URL: %v", err)
```
hwh33: Why crash the application for this?

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+
+	detour.ForceWhitelist(u.Host)
+
+	userId := strconv.Itoa(int(opts.UserConfig.GetUserID()))
+	_configService.opts = opts
+	_configService.clientInfo = &apipb.ConfigRequest_ClientInfo{
+		FlashlightVersion: common.LibraryVersion,
+		ClientVersion:     common.CompileTimeApplicationVersion,
+		UserId:            userId,
+		ProToken:          opts.UserConfig.GetToken(),
```
hwh33, line 105: What happens if we fetch a new pro token?  It looks like this is never updated.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	close(cs.done)
+	cs.running = false
+}
+
+// fetchConfig fetches the current config from the server and updates the client's config if a change
+// has occurred. It returns the extra sleep time received from the server response and any error that
+// occurred.
+func (cs *configService) fetchConfig() (int64, error) {
```
hwh33, line 139: This function needs concurrency protection (`cs.mu.Lock()`)

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+		cs.opts.RoundTripper,
+		cs.opts.UserConfig,
+		cs.logger,
+	)
+	if err != nil {
+		return nil, 0, fmt.Errorf("config request failed: %w", err)
+	}
+	defer resp.Close()
+
+	gzReader, err := gzip.NewReader(resp)
```
hwh33: Are protocol buffer responses from the back-end gzipped?  That'd be a bit odd I think, but I guess I don't actually know.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return nil, sleep, fmt.Errorf("unable to open gzip reader: %w", err)
+	}
+
+	configBytes, err := io.ReadAll(gzReader)
```
hwh33: This error is never checked.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	newConf := &apipb.ConfigResponse{}
+	if err = proto.Unmarshal(configBytes, newConf); err != nil {
+		return nil, 0, fmt.Errorf("unable to unmarshal config: %w", err)
+	}
+
+	return newConf, sleep, err
+}
+
+// newRequest returns a new ConfigRequest with the current client info, proxy ids, and the last
+// time the config
```
hwh33: This comment is incomplete.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+func (cs *configService) newRequest() *apipb.ConfigRequest {
+	conf := cs.configHandler.GetConfig()
+	proxies := []*apipb.ProxyConnectConfig{}
+	if conf != nil { // not the first request
+		proxies = conf.GetProxy().GetProxies()
+	}
+
+	ids := make([]string, len(proxies))
+	for _, proxy := range proxies {
+		ids = append(ids, proxy.GetTrack())
```
hwh33: This is not the proxy ID.  You want `proxy.Name`.

services/config.go
```diff
@@ -0,0 +1,264 @@
...
+	return confReq
+}
+
+// configIsNew returns true if country, proToken, or ip in currInfo differ from new or if new has
+// proxy configs.
+func configIsNew(cur, new *apipb.ConfigResponse) bool {
+	return cur.GetCountry() != new.GetCountry() ||
+		cur.GetProToken() != new.GetProToken() ||
+		len(new.GetProxy().GetProxies()) > 0
+}
```
hwh33: Could you add a test to future-proof this?  If new fields are added to `apipb.ConfigResponse`, they will be missed by this function.  There's no easy way to alert maintainers to the need to update this function without some kind of automated failure.

It won't be easy.  Maybe some kind of reflection tricks to populate every field and test that this function can catch changes to each?

hwh33: There might be existing solutions for this kind of thing.

services/http.go
```diff
@@ -0,0 +1,58 @@
...
+	logger golog.Logger,
+) (io.ReadCloser, int64, error) {
+	req, err := http.NewRequest("POST", originURL, buf)
+	if err != nil {
+		return nil, 0, fmt.Errorf("unable to create request for %s: %w", originURL, err)
+	}
+
+	common.AddCommonHeaders(user, req)
+	req.Header.Set("Content-Type", "application/x-protobuf")
+	req.Header.Set("Accept", "application/x-gzip")
```
hwh33: I see this is where the gzip assumption starts.  Again, does this back-end respect this header for "application/x-protobuf"?  Protocol buffer messages are usually dense, so I don't think it's typical to apply compression.

services/http.go
```diff
@@ -0,0 +1,58 @@
...
+	resp, err := rt.RoundTrip(req)
+	if err != nil {
+		return nil, 0, fmt.Errorf("request to %s failed: %w", originURL, err)
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, 0, fmt.Errorf("bad response code: %v", resp.StatusCode)
+	}
+
+	// TODO: do we need to log all the response headers?
```
hwh33: Heh, good question.  Did this come from existing code?  Doesn't hurt I guess, especially at the debug level.

services/http.go
```diff
@@ -0,0 +1,58 @@
...
+	logger.Debugf("Response headers from %v:\n%v", originURL, resp.Header)
+
+	var sleepTime int64
+	sleepVal := resp.Header.Get(common.SleepHeader)
+	if sleepVal != "" {
+		if sleepTime, err = strconv.ParseInt(sleepVal, 10, 64); err != nil {
+			logger.Errorf("Could not parse sleep val: %v", err)
```
hwh33: Let's set a default sleep time in this case.  How's one hour sound to you?

hwh33 (commented):
Nice work so far @garmr-ulfr!

> The config request service utilizes the callRandomly code from bypass that adds jitter between requests to avoid creating a pattern that censors could detect.

Just FYI, another reason is to prevent [the thundering herd problem](https://encore.dev/blog/thundering-herd-problem).  Probably the more important reason IMO.

hwh33:
> Just FYI, another reason is to prevent [the thundering herd problem](https://encore.dev/blog/thundering-herd-problem). Probably the more important reason IMO.

On that note, if we're generalizing the fetching code, I think we should add exponential back-off on error.  Let's max that out at 10 minutes (not including jitter).

config/config_test.go
garmr-ulfr: These will be deleted here and replaced with new tests located with their corresponding code (`services`, `config/proxy`). I left them there as reference for me but commented them out so the CI wouldn't run them when I pushed commits.

config/initializer.go
garmr-ulfr: Oh, missed these. These will be deleted.

config/initializer.go
garmr-ulfr: Yup, also will be deleted. All the code I commented out will be deleted and was just left as reference for me to make sure I kept the same values/functionality. Thanks for catching these! I'm going to remove all these now.

config/proxy/proxy.go
garmr-ulfr: Yeah, I know. I originally was going to use `ClientConfig` but that's already used in `config/client_config.go` and I couldn't think of anything else at the moment. I figured once we include the global config we'd be able to rename it to just `Config` since it wouldn't clash. 

In case you haven't noticed by now, I'm not so great at naming :laughing:. So I'm open to any and all suggestions.

config/proxy/proxy.go
garmr-ulfr: Oh, good point. I'll update it!

config/proxy/proxy.go
garmr-ulfr: I thought about that, but I wanted to hide the use of `eventual` from callers as they don't need to know internals. I have no problem changing it though if you think it would be better to accept a context instead.

config/proxy/proxy.go
garmr-ulfr: They were just for convenience. I was on the fence about including them or not anyway. I'll remove them.

config/proxy/proxy.go
garmr-ulfr: This is how it was in the original code so I left it. I think it treats an empty file as if the file didn't exist. Maybe it would be better to return an empty `ProxyConfig` instead if the file is empty?

flashlight.go
garmr-ulfr: That is correct. I don't know why I thought it would just be the zero value of a bool.

flashlight.go
garmr-ulfr: `proxyconfig.notifyListeners` is only called when there is a new config, so when `fn` is called in `proxyconfig` `new` is guaranteed to not be `nil`. However, that being said, it doesn't prevent callers of `proxyconfig.OnConfigChange` from calling `fn` themselves without passing `nil` for `new`, so I will have to fix that at line 503.

flashlight.go
garmr-ulfr: Yeah, I'll update this to check for nil.

geolookup/geolookup.go
garmr-ulfr: Yup, I do. I'll update it.

services/service.go
garmr-ulfr: Hmmm..I guess auto-import doesn't prefer v2 over v1..

services/config.go
garmr-ulfr: > Not sure if this comment was ported from somewhere else, but I find the near double negative confusing.
Yeah, that is the original comment. 

> Actually, this guarantee is not possible with your current API. You don't have any control in this package over the first result of ConfigHandler.GetConfig (since it's an interface with no strict requirements on the source of the first call).

If `Sticky` is true, `configService` returns [here](https://github.com/getlantern/flashlight/pull/1395/files/f7de44c45aecd7155b6b96f0ae9c1fbdacfd22af#diff-6edc66a61226196eb1a3035bd1dad77b4c22d2cab11dc33ae62c9cfa4c71a8e9R137-R139) and doesn't start the loop of fetching the remote config. `configService` only fetches remote configs and be an interface between `ConfigHandler` and other packages. Since `ConfigHandler` is responsible for the actual config, `configService` assumes that the initial value has already be set. `proxyconfig.Init` reads in the local config file and sets it as the initial value for `proxyconfig.config` [here](https://github.com/getlantern/flashlight/pull/1395/files/f7de44c45aecd7155b6b96f0ae9c1fbdacfd22af#diff-be33908ff3493e3cab2744eb648322f15e8af73624cac4cbac593caa1473f24bR59-R64).

Maybe renaming `Sticky` to something like `FetchRemote` would make it much more clear.

Also, initially, I considered implementing it such that `configService`'s only purpose was to fetch the remote config and update it as needed and have other packages call `proxyconfig` directly to get the config (since this is the only interaction between other packages and `configService`). This way, if `Sticky` is true then you just don't start `configService`. I don't remember why I decided not to go that route, but maybe that is the solution. What do you think?

services/config.go
garmr-ulfr: So just set a fixed value?

services/config.go
garmr-ulfr: I don't know. That's directly from the current [implementation](https://github.com/getlantern/flashlight/blob/main/config/fetcher.go#L54-L58).

services/config.go
garmr-ulfr: It's updated in `updateClientInfo` called [here](https://github.com/getlantern/flashlight/pull/1395/files/f7de44c45aecd7155b6b96f0ae9c1fbdacfd22af#diff-6edc66a61226196eb1a3035bd1dad77b4c22d2cab11dc33ae62c9cfa4c71a8e9R190-R193).

services/config.go
garmr-ulfr: Why would that be odd? Are they not usually compressed? I would assume everything (except maybe single TCP request/response) would be gzipped since it reduces bandwidth. I don't actually know what the text form of a protocol buffer looks like, so maybe it's already very small.

services/config.go
garmr-ulfr: I can go look at how `testify/assert` does it.

services/http.go
garmr-ulfr: Again, I just assume it's compressed. I'll look at `lantern-cloud` and check if it is.

services/http.go
garmr-ulfr: Yeah, comes from the existing code.

services/http.go
garmr-ulfr: `sleepVal` here is extra sleep time. If we wanted an hour between polls, wouldn't that be `PollInterval`?

services/config.go
garmr-ulfr: Forgot about `reflect.DeepEqual`..it's actually very easy.

services/config.go
garmr-ulfr: Are you sure? The only possible race conditions I see are:
* `cs.lastFetched = time.Now()` => `lastFetched` is only access by this function and `newRequest` which would be on the same call stack.
* `cs.updateClientInfo(newConf)` => `clientInfo` is also only accessed by this function and `newRequest`. (even with removing `updateClientInfo` and updating it here directly)
* ~`cs.configHandler.SetConfig(newConf)` => `configHandler` handles concurrency itself using `eventual`.~

The function itself has to return before it's called again. Am I missing something??

flashlight.go
garmr-ulfr: Wait, actually, this won't panic. Because I'm using the test form (with the optional second return value) it doesn't panic. It sets `b` to false (zero value of the type I'm trying to assert to) and `ok` (which I toss) to false because it failed.

https://go.dev/play/p/R7tPQiqjYA1

flashlight.go
garmr-ulfr: I remember now. This will set `obfuscate` to false if `flagsAsMap["readableconfig"]` returns `nil`. See comment below.

apipb/legacy.go
garmr-ulfr: That is correct! I haven't delved past `StartBackgroundServices` in `flashlight.go` so I don't exactly know how deep the legacy config format goes. I assume it's used in a plethora of places, but I hope I'm wrong.

services/config.go
garmr-ulfr: > ..it's actually very easy.

Maybe not..I'd need to account for optional fields in `new`..I'll figure it out.

services/service.go
garmr-ulfr: I updated it to subtract it and clip negatives to zero.

services/config.go
garmr-ulfr: @hwh33 and I discussed this on a call and decided that it would be better to move the responsibility of returning the config to other packages to `proxyconfig`, resulting in `configService`'s only responsibility is to fetch new configs form the server and relaying that to `proxyconfig` accordingly.

services/config.go
garmr-ulfr: It might be easier to have the server return a different status code, like 201, if there are no changes. This way we wouldn't have to update `configIsNew` every time new fields are add or removed. The server already checks if new values needs to be sent back to the client, there's no need for the client to do all the checks again.

services/config.go
myleshorton: The gzipping was from when responses were returned as json/yaml, but I'm unclear if gzipping makes proto smaller or not.

garmr-ulfr: Alright. Well, I'll just remove it for now. We can always add it back in later if we want.

services/http.go
garmr-ulfr: Removing gzip compression.

services/config.go
garmr-ulfr: Discussed on a call. We couldn't find a race condition and the only need for `cs.mu` is to protect `cs.running`. I will change the mutex name and add a comment to make that clear.

flashlight.go
garmr-ulfr: `startConfigService` now checks for nil before calling `fn`.

services/http.go
```diff
@@ -0,0 +1,104 @@
...
+	originURL string,
+	buf io.Reader,
+	rt http.RoundTripper,
+	user common.UserConfig,
+) (*http.Response, int64, error) {
+	resp, err := s.doPost(originURL, buf, rt, user)
+	if err == nil {
+		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent {
+			err = fmt.Errorf("bad response code: %v", resp.StatusCode)
+			goto backoff
```
WendelHime: This is a optional suggestion since it's comprehensible and not complex case use of `goto`: We could replace this `goto` creating and calling a `retry` function (I'm assuming it's  going to be retried later), something like this:
```
func (s *sender) retry(err error) (*http.Response, int64, error) {
	if s.atMaxRetryWait {
		// we've already reached the max wait time, so we don't need to perform the calculation again.
		// we'll still increment the fail count to keep track of the number of failures
		s.failCount++
		return nil, int64(maxRetryWait.Seconds()), err
	}

	wait := time.Duration(math.Pow(2, float64(s.failCount))) * retryWaitSeconds
	s.failCount++

	if wait > maxRetryWait {
		s.atMaxRetryWait = true
		return nil, int64(maxRetryWait.Seconds()), err
	}

	return nil, int64(wait.Seconds()), err
}
```
And replace line 42 and 58-74 by `s.retry(err)`. We could also use a library to handle retries like https://github.com/avast/retry-go or https://github.com/sethvargo/go-retry

WendelHime (approved):
The PR looks really good! I've added a small optional suggestion because I'm allergic to goto and labels but it can be ignored!

services/config.go
```diff
@@ -0,0 +1,241 @@
...
+
+// ConfigHandler handles updating and retrieving the client config.
+type ConfigHandler interface {
+	// GetConfig returns the current client config.
+	GetConfig() *apipb.ConfigResponse
+	// SetConfig sets the client config to the given config.
+	SetConfig(new *apipb.ConfigResponse)
+}
+
+var _configService = &configService{sender: &sender{}}
```
WendelHime, line 64: It looks like all references from `_configService` are only contained at the `StartConfigService`, can we move it to inside the function, initialize there and I guess we could call `cfgService` or just `service` or `cs` (which is being used at other functions)

WendelHime: Otherwise this variable will be initialized when the application is starting or loading the services package, if this is the intention it's ok but I guess it isn't since isn't referred outside the package

geolookup/geolookup.go
```diff
@@ -2,261 +2,83 @@ package geolookup
...
-	refreshRequest = make(chan interface{}, 1)
-	currentGeoInfo = eventual.NewValue()
-	watchers       []chan bool
-	persistToFile  string
-	mx             sync.Mutex
-	roundTripper   http.RoundTripper
-)
+	// _country and _ip are eventual values that hold the current country and IP as strings.
+	_country = eventual.NewValue()
+	_ip      = eventual.NewValue()
```
WendelHime: There's a reason for the `_` on the name? I guess we could call them country and ip directly or add a eventual prefix

flashlight.go
```diff
@@ -437,46 +437,140 @@ func (f *Flashlight) Run(httpProxyAddr, socksProxyAddr string,
...
 	onError func(err error),
 ) {
-	stop := f.StartBackgroundServices()
+	// Until we know our country, default to IR which has all detection rules
+	log.Debug("Defaulting detour country to IR until real country is known")
+	detour.SetCountry("IR")
+
+	stop, err := f.StartBackgroundServices()
+	if err != nil {
+		log.Errorf("Unable to start background services: %v", err)
```
WendelHime:
```suggestion
		log.Errorf("Unable to start background services: %w", err)
```

flashlight.go
```diff
@@ -437,46 +437,140 @@ func (f *Flashlight) Run(httpProxyAddr, socksProxyAddr string,
...
+// is temporary until all code is updated to use the new config format.
+//
+// TODO: Update all code to use the new config format and remove this method.
+func (f *Flashlight) convertNewProxyConfToOld(proxies []*apipb.ProxyConnectConfig) map[string]*commonconfig.ProxyConfig {
+	proxyMap := make(map[string]*commonconfig.ProxyConfig)
+	for _, p := range proxies {
+		// use lantern-cloud to map to legacy so we don't have to write it ourselves
+		lc, err := apipb.ProxyToLegacyConfig(p)
+		if err != nil {
+			log.Errorf("Unable to convert %s proxy config to legacy: %v", p.Name, err)
```
WendelHime, line 564:
```suggestion
			log.Errorf("Unable to convert %s proxy config to legacy: %w", p.Name, err)
```

services/bypass.go
```diff
@@ -0,0 +1,279 @@
...
+		if !unsupportedTransports[v.PluggableTransport] {
+			supportedInfos[k] = v
+		}
+	}
+
+	dialers := chained.CreateDialersMap(configDir, supportedInfos, userConfig)
+	for k, v := range supportedInfos {
+		dialer := dialers[k]
+		if dialer == nil {
+			logger.Errorf("bypass: no dialer for %v", k)
```
WendelHime, line 111:
```suggestion
			logger.Errorf("bypass: no dialer for %w", k)
```

WendelHime: Wait this is not an error

config/proxy/proxy.go
```diff
@@ -0,0 +1,210 @@
...
+			// proxy config. So we want to overwrite the existing proxy configs with the new ones.
+			updated.Proxy = nil
+		}
+
+		proto.Merge(updated, new)
+	}
+
+	c.config.Set(updated)
+	if err := saveConfig(c.filePath, updated, c.obfuscate); err != nil {
+		log.Errorf("Failed to save client config: %v", err)
```
WendelHime:
```suggestion
		log.Errorf("Failed to save client config: %w", err)
```

geolookup/geolookup.go
garmr-ulfr: There used to be. I have since changed some things and I forgot they were named that way. I'm going to change them.

services/config.go
garmr-ulfr: This so `_configService` is a singleton to prevent multiple instances of it being started. 

> Otherwise this variable will be initialized when the application is starting or loading the services package, if this is the intention it's ok but I guess it isn't since isn't referred outside the package

This is fine since `StartConfigService` is going to be called during startup. I intentionally initialized it so there wouldn't have to be a mutex that would have to be locked/unlocked just to check if it was null and then initialized it.

flashlight.go
garmr-ulfr: Since we're just logging the error here, we want it to print nicely. Otherwise, if we were returning it then, yes, we would wrap it.

flashlight.go
garmr-ulfr: Same as above.

config/proxy/proxy.go
garmr-ulfr: Same as above.

config/proxy/proxy.go
hwh33: Hmm yeah it'd be nice if `config` weren't already taken!

Let's see.  Looking at the API, this package facilitates watching for config updates.  So maybe `cfgwatch`?  I don't totally love it and I'd still prefer `config` to be honest - but that's taken!

config/proxy/proxy.go
hwh33: > but I wanted to hide the use of eventual from callers as they don't need to know internals

I agree with that!  But `context.Context` is the part being exposed, and that's a prominent part of the standard library.  I would change it to take a context, yeah.

config/proxy/proxy_test.go
```diff
@@ -0,0 +1,148 @@
...
+	withTempConfigFile(t, conf, false, func(tmpfile *os.File) {
+		Init("", false)
+		existing, _ := GetConfig(0)
+
+		want := fmt.Sprintf("%+v", conf)
+		got := fmt.Sprintf("%+v", existing)
+		assert.Equal(t, want, got, "failed to read existing config file")
+	})
+
+	resetConfig()
```
hwh33: Nit: I would make this a deferred call at the beginning of the test.  Since you're using `assert`, there's no existing issue.  However, if any `require` checks are added, this test could exit before `resetConfig` is called, breaking other tests in unexpected ways.

flashlight.go
```diff
@@ -246,28 +247,21 @@ func (f *Flashlight) notifyProxyListeners(proxies map[string]*commonconfig.Proxy
 	}
 }
 
-func (f *Flashlight) startConfigFetch() func() {
-	proxiesDispatch := func(conf interface{}, src config.Source) {
-		proxyMap := conf.(map[string]*commonconfig.ProxyConfig)
-		f.notifyProxyListeners(proxyMap, src)
-	}
+func (f *Flashlight) startGlobalConfigFetch() func() {
+	// TODO: update global config fetching to use new config service
```
hwh33: I don't think we need this TODO here.  If you want to add it, I would prefer to see it more descriptive as far as what needs to be done (the global config needs to first become part of the consolidated config response).

services/http.go
garmr-ulfr: > This is a optional suggestion since it's comprehensible and not complex case use of goto

Yeah, I need to change that. 

> We could also use a library to handle retries like https://github.com/avast/retry-go or https://github.com/sethvargo/go-retry

I prefer not to use external libraries for small, simple things like this. There's less dependencies.

flashlight.go
hwh33: Wow yeah, you can cast the `nil` value to anything here:
https://go.dev/play/p/s7uMJRj7Bl6

Interesting.

services/config.go
hwh33: Yep

services/config.go
hwh33: Heh, fascinating.  Let's change that to a `logger.Errorf`

services/config.go
hwh33: I think `updateClientInfo` is gone, but it looks like this is now updated in `fetchConfig`.

services/http.go
```diff
@@ -0,0 +1,104 @@
...
+
+	wait := time.Duration(math.Pow(2, float64(s.failCount))) * retryWaitSeconds
+	s.failCount++
+
+	if wait > maxRetryWait {
+		s.atMaxRetryWait = true
+		return nil, int64(maxRetryWait.Seconds()), err
+	}
+
+	return nil, int64(wait.Seconds()), err
```
hwh33: (I support @WendelHime's suggestion to just make this a helper function.  The following still applies.)

You can simplify this to:
```
	wait := time.Duration(math.Pow(2, float64(s.failCount))) * retryWaitSeconds
	s.failCount++

	if wait > maxRetryWait {
		wait = maxRetryWait
	}

	return nil, int64(wait.Seconds()), err
```

and get rid of `s.atMaxRetryWait`.  The extra calculation referenced in the comment is not significant.

services/http.go
```diff
@@ -0,0 +1,104 @@
...
+) (*http.Response, error) {
+	req, err := http.NewRequest("POST", originURL, buf)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create request for %s: %w", originURL, err)
+	}
+
+	common.AddCommonHeaders(user, req)
+	req.Header.Set("Content-Type", "application/x-protobuf")
+	// Prevents intermediate nodes (domain-fronters) from caching the content
+	req.Header.Set("Cache-Control", "no-cache")
```
hwh33: It's odd for this abstracted type to be handling specific headers like this.  Could `doPost` and `post` just take an `http.Request` and let the caller handle this stuff?

services/http.go
garmr-ulfr: Already updated, just not pushed yet.

config/proxy/proxy_test.go
garmr-ulfr: Good point. I'll update it.

services/http.go
garmr-ulfr: Updated as suggested.

services/config.go
garmr-ulfr: `jitter` is now a constant in `service`. However, created a new function, `callRandonlyWithJitter`, that accepts a `jitter` and `callRandomly` just calls it. This is so we can still use the tests I already wrote without having to modify them.

config/proxy/proxy.go
garmr-ulfr: `ProxyConfig` has been renamed to `UserConfig`, as well as package name, etc., since it stores the user's location, IP, and the proxies assigned to them. Maybe, one day, we'll be able to use `config`..

services/config.go
garmr-ulfr: The server now returns `http.StatusNoContent` if there are no updates to the config.

hwh33 (approved):
LGTM.  Awesome work @garmr-ulfr!

Let's open a PR into lantern-client with these changes.  We'll try to get a build out and prioritize thorough testing.

integrationtest/integrationtest.go
```diff
@@ -67,20 +68,44 @@ type Helper struct {
...
+	ShadowsocksProxyServerPort    int32
+	ShadowsocksmuxProxyServerPort int32
+	TLSMasqProxyServerPort        int32
+	HTTPSSmuxProxyServerPort      int32
+	HTTPSPsmuxProxyServerPort     int32
+	HTTPServerPort                int32
+	HTTPSServerPort               int32
+
+	// Deprecated: Use BaseServerAddr and HTTPSProxyServerPort instead.
+	HTTPSProxyServerAddr string
```
hwh33: Since this is a package used (a) exclusively for tests and (b) only by our own projects, you don't need to deprecate these.  You can just make a breaking change to the API and synchronize that change when you update downstream projects like `lantern-client`.

hwh33 (approved):
One minor comment, but LGTM otherwise!

garmr-ulfr:
There is bug causes the fetch to fail  when starting Lantern and there are existing proxy configs with IPs that have rotated out.

garmr-ulfr:
> There is bug causes the fetch to fail when starting Lantern and there are existing proxy configs with IPs that have rotated out.

Fixed.

PR #1396, opened 2024-07-17
Adding SNI config support

WendelHime:
This PR allow flashlight to load FrontingSNIs from the global configuration and send them to the fronted library so we can fill SNIs when sending requests. This PR requires getlantern/fronted#40 to be merged before, so we can update the  `go.mod` reference


embeddedconfig/global.yaml
```diff
@@ -255,6 +255,18 @@ client:
...
           rejectstatus: [403]
+        frontingsnis:
+          default:
+            usearbitrarysnis: false
+            arbitrarysnis:
+            - amazon.com
+            - mercadopago.com
+            - facebook.com
+            - instagram.com
+            - twitter.com
```
hwh33: Let's use the domains from [this list](https://github.com/getlantern/lantern-cloud/blob/175729685c665e2ae92e1c4e458d8fa2cc5d0dff/cmd/tlsmasq-tool/tmasqorigins/default.yaml) (I realize the ones you've got here now are just placeholders).

config/client_config.go
```diff
@@ -80,15 +82,24 @@ func NewClientConfig() *ClientConfig {
...
 func (c *ClientConfig) FrontedProviders() map[string]*fronted.Provider {
+	sniRegion := geolookup.GetCountry(0)
+	if sniRegion == "" {
+		sniRegion = "default"
+	}
 
 	providers := make(map[string]*fronted.Provider)
 	for pid, p := range c.Fronted.Providers {
+		sniConfig := p.FrontingSNIs[sniRegion]
+		if sniConfig != nil && sniConfig.UseArbitrarySNIs && len(sniConfig.ArbitrarySNIs) == 0 {
```
hwh33: We want to fall back to the default SNI config if `sniRegion` does not appear in `p.FrontingSNIs`.  Currently, if there is no entry in `p.FrontingSNIs` for `sniRegion`, then `sniConfig` will never be populated (`sniConfig.UseArbitrarySNIs` will be ignored because `sniConfig != nil` will fail).

I think we instead want something like:
```
sniConfig, ok := p.FrontingSNIs[sniRegion]
if !ok {
  sniConfig = p.FrontingSNIs["default"]
}
```

Or am I misunderstanding something?

WendelHime: aaaa I have misunderstood, I thought we only want to use the default arbitrary list when the region has an empty arbitrary list, will fix that soon

WendelHime: Wait I didn't misunderstood, the provided example does mention that, if there are no arbitrary sni provided and we want to use it, we will use it from the default section (see cn example and comment):

```
fronting-snis:
  default:
    use-arbitrary-snis: false
    arbitrary-snis:
      - example.com
      - example2.com
      - ... etc ...
  ir:
    use-arbitrary-snis: true
    arbirtary-snis:
      - iran-example.com
      - iran-example2.com
      - ... etc ...
  cn:
    use-abritrary-snis: true
    # no abritrary-snis list; use list from default section
```

hwh33: The example looks good and your updated code implements it correctly!

hwh33 (approved):
LGTM!  Nice work @WendelHime!

PR #1398, opened 2024-08-07
Removing sampling from autoupdate spans

WendelHime:
This PR removes span sampling from autoupdate install and download operations. I'm doing this with the hope to get more information about autoupdate issues that are not happening as expected.


hwh33 (approved):
Just keep an eye on these metrics in Honeycomb's Usage mode, please 🙂 .  Just in case our math (in our Slack conversation) was somehow off.

PR #1400, opened 2024-08-08
Updating genconfig for using fronting SNIs

WendelHime:
This pull request removes deprecated calls from genconfig and update the structure for allowing parsing fronting SNI config from provider map.


genconfig/provider_map.yaml
```diff
@@ -45,3 +45,7221 @@ akamai:
...
+      - studopedia.ru
+      - yurticikargo.com
+      - ketabrah.ir
+      - vesti.bg
+      - parscoders.com
+    ir:
+      usearbitrarysnis: true
+      arbitrarysnis:
+    cn:
+      usearbitrarysnis: true
```
hwh33: (I know I already approved this in the last PR, but I've given it more thought)

Domain fronting is reliable in China, so let's not use arbitrary SNIs there for the moment.  Let's keep the China-specific SNI list here for future use though.  Let's use arbitrary SNIs in Iran, Myanmar, Russia, and the UAE.  I'm not sure to be honest whether it will help in all of those countries, but it'll be interesting to see!

Here's some data (from Honeycomb) on domain fronting error rates over the last week.  I'm basing the decisions above on this:

image

And here's the query if you want to play around with it yourself:

<details>
  <summary>Query</summary>

image

</details>

hwh33 (approved):
Looks good to me!  Just a few suggestions on which countries we should target with this.

hwh33:
Feel free to merge once those changes are in!  Don't worry about those two failing tests for getting this in - they're clearly unrelated and we can fix them separately.

PR #1402, opened 2024-08-12
removing duplicated routing rules

WendelHime:


embeddedconfig/global.yaml.tmpl
```diff
@@ -284,7 +284,6 @@ nameddomainroutingrules:
 domainroutingrules:
   huya.com: d
   douyucdn2.cn: d
-  apple.com: d
```
hwh33: Let's instead remove the "must-direct" rule and leave this one in place.  IMU, that will allow the client to eventually fall back to proxying if they're unable to reach apple.com.

hwh33: You can go ahead and merge after this change.

PR #1410, opened 2024-08-27
Add auth flag

jigar-f:


reflog (approved):

jigar-f:
@atavism I have a question here, Since we uploaded the Android build,  If we enable auth flow here, Di think that will cause an issue on the build that is already live?  Since we have some UI parts already on there.

myleshorton:
Do you still need this @jigar-f?

jigar-f:
We can close it, At the moment we have auth enabled for ios only

PR #1414, opened 2024-09-23
Removing cloudfront test verification

WendelHime:
Since cloudfront DDF was disabled the github action that generates automatically/fetch the global config is broken. This PR removes some cloudfront verifications that should allow it to run as expected.


garmr-ulfr (approved):
LGTM. Perfect timing, I was just looking at test logs to see why it was failing.

hwh33 (approved):
Nice, thanks for taking care of this!

PR #1415, opened 2024-09-23
Simple dialer selector

garmr-ulfr:
This replaces the MAB with a simple selector for choosing which dialer to use. The selector just loops through the list of dialers and chooses the first one to connect. When dialing, the selector will always start with the most recent dialer that succeeded.


chained/selector_test.go
```diff
@@ -0,0 +1,134 @@
...
+			&mockDialer{name: "Microsoft still"},
+		},
+	)
+	_, err := selector.Dial("", "working windows updates")
+	assert.Error(t, err)
+}
+
+const selectorTestAddr = "http://test.addr"
+
+type mockDialer struct {
```
WendelHime, line 81: This suggestion is purely optional and I'm just adding this solely for discussion purposes. Can we use https://github.com/uber-go/mock for generating interfaces mocks? It's generally easier to maintain and generate them whenever we need to update them (using `go generate ./...`). But again, this can be ignored, the current solution woks and it's good.

WendelHime (approved):
LGTM! I've  just added one small comment for discussion about mock tools but it can be ignored.

chained/selector_test.go
garmr-ulfr: I'm not really familiar with it so I'll have to check it out. But, if it makes creating mocks easier, then I'm all for it.

myleshorton:
I again don't think this is a good idea at all unless we have logic on the backend to actually support sending the user proxies in an informed way, as that was creating a terrible user experience.

myleshorton (changes requested):
I think the notion of having the client just do whatever the server says is generally OK, but in the absence of the server being able to tell the client what to use in a meaningful way, we're back to where we were prior to the MAB being added, which was a disaster for users, and which is why we implemented MAB. 

As I previously mentioned, without those changes on the server side, I would be completely opposed to this seeing the light of day. I do think it would be a good idea to be able to tell the client "just do what I tell you", so we could add some sort of flag in the config response indicating that. That is actually necessary for when we implement the server side logic for optimizing the protocols and data centers to use for each client (loosely server-side MAB, but more generally just server side optimization of proxy assignment using some form of machine learning).

garmr-ulfr:
> I again don't think this is a good idea at all unless we have logic on the backend to actually support sending the user proxies in an informed way, as that was creating a terrible user experience.

@myleshorton, I agree, the backend should definitely assign proxies intelligently. From my understanding, this is meant to be temporary, to help stabilize connections, until we implement something on the backend. 

> but more generally just server side optimization of proxy assignment using some form of machine learning

[Here's](https://github.com/getlantern/engineering/issues/1328#issuecomment-2292294396) one approach that utilizes DRL.

Crosse:
At the risk of wading into a burgeoning holy war 😄, might I suggest a bake-off? Don't remove MAB and keep it as the default, but add this selector and potentially the DRL one behind feature flags that we can twiddle in the global config. That way we can test the selectors in a less risky way. (Some changes to the server's logic were mentioned in a previous comment; if we go that route, then the API could apply the feature flag and inform the client at request time which selector it should use.)

garmr-ulfr:
> Don't remove MAB and keep it as the default, but add this selector and potentially the DRL one behind feature flags that we can twiddle in the global config. That way we can test the selectors in a less risky way.

That's actually a really good idea. @myleshorton, @hwh33, what do you guys think? If you agree, I'll go ahead and get started on implementing this.

myleshorton:
Agreed! 

That's what I was saying with:

> I do think it would be a good idea to be able to tell the client "just do what I tell you", so we could add some sort of flag in the config response indicating that.

I don't feel strongly about it being in the global config vs the config server, although I do maybe slightly prefer the config server just given that it's already a config for this specific thing. That would also make this much more flexible, as we could use the feature flagging in the API to enable this for specific segments of users.

myleshorton:
I'm not exactly sure on the specific algorithm to use, but I do think it might make sense to use some of Google's stuff just given that we already have all the data in BigQuery (if it's not prohibitively expensive). This looks potentially promising:

https://cloud.google.com/blog/topics/developers-practitioners/build-reinforcement-learning-recommendation-application-using-vertex-ai

hwh33:
> might I suggest a bake-off? Don't remove MAB and keep it as the default, but add this selector and potentially the DRL one behind feature flags that we can twiddle in the global config

That's a great idea!  I wonder how we will gather the results of this experiment, but we can discuss that elsewhere (I'll start a thread on Slack).

Let's use the global config since the client already knows how to read feature flagging out of the global config (and we can still enable the feature for specific segments of users).

@garmr-ulfr I'm going to close this PR and ask you to open two new ones:
1. A first one which simply adds the simple selector, but leaves the bandit dialer in place.  We won't use the simple selector in this PR, it'll just be inert code, but it should implement the same API as the bandit dialer so that we could in theory switch one out for the other.
2. A second PR which allows us to enable the simple selector via the global config's [feature flagging capabilities](https://github.com/getlantern/flashlight/blob/160434473c8f4b799a1a6f2067b39cde6786f385/config/global.go#L57-L62).  The bandit dialer should remain the default.

PR #1418, opened 2024-09-26
fix genconfig test

WendelHime:
TLDR: We should be able to test the genconfig logic without using external IPs

This pull request aims to mock genconfig HTTP requests so we can only test the logic for generating the configuration correctly. Currently there's two functions that need to make some requests for retrieving certs and verifying if the given masquerade IPs are working as expected, those functions were mocked and now we can test the logic successfully without having to interact with external IPs


garmr-ulfr (approved):
LGTM. Still need to check out `gomock`. It looks very simple and similar to [gock](https://github.com/h2non/gock)

PR #1422, opened 2024-10-01
add simple proxy selector

garmr-ulfr:
This will add a simple selector for choosing which dialer to use. The selector just loops through the list of dialers and chooses the first one to connect. When dialing, the selector will always start with the most recent dialer that succeeded.

The simple selector will be enabled in a future PR.

### DIFF
--- a/chained/selector.go
+++ b/chained/selector.go
@@ -1,0 +1,214 @@
+package chained
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"iter"
+	"net"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	// NetworkConnect is a pseudo network name to instruct the dialer to establish
+	// a CONNECT tunnel to the proxy.
+	NetworkConnect = "connect"
+	// NetworkPersistent is a pseudo network name to instruct the dialer to
+	// signal the proxy to establish a persistent HTTP connection over which
+	// one or more HTTP requests can be sent directly.
+	NetworkPersistent = "persistent"
+
+	// maxDialerFails is the max consecutive failed dials/requests before moving to the next dialer
+	maxDialerFails = 4
+)
+
+// Dialer provides the ability to dial a proxy
+type Dialer interface {
+
+	// DialProxy dials the proxy but does not yet dial the origin.
+	DialProxy(ctx context.Context) (net.Conn, error)
+
+	// SupportsAddr indicates whether this Dialer supports the given addr. If it does not, the
+	// balancer will not attempt to dial that addr with this Dialer.
+	SupportsAddr(network, addr string) bool
+
+	// DialContext dials out to the given origin. failedUpstream indicates whether
+	// this was an upstream error (as opposed to errors connecting to the proxy).
+	DialContext(ctx context.Context, network, addr string) (conn net.Conn, failedUpstream bool, err error)
+
+	// Name returns the name for this Dialer
+	Name() string
+
+	// Label returns a label for this Dialer (includes Name plus more).
+	Label() string
+
+	// JustifiedLabel is like Label() but with elements justified for line-by
+	// -line display.
+	JustifiedLabel() string
+
+	// Location returns the country code, country name and city name of the
+	// dialer, in this order.
+	Location() (string, string, string)
+
+	// Protocol returns a string representation of the protocol used by this
+	// Dialer.
+	Protocol() string
+
+	// Addr returns the address for this Dialer
+	Addr() string
+
+	// Trusted indicates whether or not this dialer is trusted
+	Trusted() bool
+
+	// NumPreconnecting returns the number of pending preconnect requests.
+	NumPreconnecting() int
+
+	// NumPreconnected returns the number of preconnected connections.
+	NumPreconnected() int
+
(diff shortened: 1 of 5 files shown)


chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+	// signal the proxy to establish a persistent HTTP connection over which
+	// one or more HTTP requests can be sent directly.
+	NetworkPersistent = "persistent"
+
+	// maxDialerFails is the max consecutive failed dials/requests before moving to the next dialer
+	maxDialerFails = 4
+)
+
+// Dialer provides the ability to dial a proxy
+type Dialer interface {
```
hwh33: Could you reduce this interface to strictly what is needed?  There are a lot of unused functions here.  I realize that this interface predates your work here, but keeping it minimal in this context will reduce the load for future maintenance.

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		log.Errorf("dialer %s failed to dial %s %s: %v", dlr.Label(), network, addr, err)
+	}
+
+	return nil, fmt.Errorf("failed to dial %s %s", network, addr)
+}
+
+func (s *Selector) dialWithDialer(ctx context.Context, dlr Dialer, network, addr string) (net.Conn, error) {
+	for {
```
hwh33: Why are we trying the same dialer repeatedly in a loop...?

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+		if err == nil {
+			return conn, nil
+		}
+
+		select {
+		case <-ctx.Done():
+			log.Debugf("dialWithDialer: %v", ctx.Err())
+			return nil, err
+		default:
+		}
```
hwh33: If we get rid of the loop here, this should no longer be necessary.  `dlr.DialContext` should return an error if the context dies.

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+
+		dlr.MarkFailure()
```
hwh33: This should depend on `failedUpstream`.

chained/selector.go
hwh33: Looking at the comment on `maxDialerFails`, it seems like this was intentional.  I think we should just move on the first time we get an error.  An exception would be if we see `failedUpstream` - that should be bubbled up to the caller, but shouldn't cause us to rotate dialers.

hwh33: That `failedUpstream` error could indicate that the site has blocked our proxy's IP.  In that case, there would be a pathological case to what I'm suggesting in which we don't rotate dialers and the user continues to experience issues connecting from the proxy to the upstream.  However, IMO that problem lives at another layer.  I don't think this dialer should concern itself with that.  The only concern here should be whether we are able to dial connect to the proxy at all.

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+
+			return nil, err
+		}
+	}
+}
+
+// MarkFailure marks the active dialer as failed. This is used to provide feedback to the Selector
+// that requests are failing. This could be due poor proxy performance, or the proxy being blocked.
+func (s *Selector) MarkFailure() {
```
hwh33: Let's leave this out unless / until we find that we actually need it.

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+	WriteStats(w io.Writer)
+}
+
+type Selector struct {
+	dialers []Dialer
+	// active is the index of the currently active dialer
+	active atomic.Int64
+}
+
+func NewSelector(dialers []Dialer) *Selector {
```
hwh33: We don't need to do this now, but just a note for the future when we integrate this: we'll want some way to retain prior knowledge I think.  One way could be something like:
```
// Update will overwrite the selectors set of dialers.
func (s *Selector) Update(dialers []Dialer)
```
where `Update` goes through `dialers` to see if its currently active dialer is in the new set.  If so, it will hang on to that as the active dialer.  If not, we can just start over.

chained/selector.go
```diff
@@ -0,0 +1,214 @@
...
+	// connections created via this dialer.
+	DataRecv() uint64
+
+	// Stop stops background processing for this Dialer.
+	Stop()
+
+	WriteStats(w io.Writer)
+}
+
+type Selector struct {
```
hwh33: Nit: this name is too general to be meaningful IMO.  Maybe `LoopDialer`?  Something like that?

chained/selector.go
garmr-ulfr: > where Update goes through dialers to see if its currently active dialer is in the new set.

Would two instances of the same dialer be considered the same or only if they're the same instance?

chained/selector.go
garmr-ulfr: Good point. Changed.

chained/selector.go
garmr-ulfr: Didn't you originally say that this was needed? 
It was removed though.

garmr-ulfr:
We've decided to scrap this and modify the [MAB to skip failing dialers](https://github.com/getlantern/flashlight/pull/1426) instead. Closing.

PR #1424, opened 2024-10-08
IOS Logs update

jigar-f:
We recently discovered that, when reporting an issue, iOS does not include tunnel logs. Since the tunnel runs in a separate process, the log paths are different. we need to ensure that iOS zips two folders: `logs` for the app and `tunnelLogs` for the tunnel process.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -31,13 +31,11 @@ var (
 	log          = golog.LoggerFor("flashlight.logging")
 	processStart = time.Now()
 
-	logFile        io.WriteCloser
-	tunnnelLogFile io.WriteCloser
-	errorPWC       io.WriteCloser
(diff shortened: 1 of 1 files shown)


jigar-f:
Do let me know if I am missing anything here or if there might be some other (better way to do this) thanks!

jay-418:
One quick side question: are these logs able to be manually retrieved like we [do for desktop](https://github.com/getlantern/customer-support/blob/main/how-to/logs.md)?

@greySkiesMarsh struggles to diagnose issues for users when the normal issue submission pipeline fails.

jigar-f:
@jay-418, unfortunately, there’s no direct way to download app group data on iOS. There isn’t an official method to do it.

garmr-ulfr (approved):
LGTM

PR #1425, opened 2024-10-08
adding support to download WASM files for WATER

WendelHime:
This PR add support to retrieve WASM file and load them before running the proxy. Currently it's using a DirectThenFronted roundtrip because, it'll always fail if there's no other proxies/dialer available. At flashlight side, it should support HTTPS URLs and magnet links, but I've only tested the HTTPS integration.


chained/mocks.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	context "context"
+	io "io"
+	reflect "reflect"
+
+	gomock "go.uber.org/mock/gomock"
+)
+
+// MockWASMDownloader is a mock of WASMDownloader interface.
+type MockWASMDownloader struct {
```
hwh33: Could you move all of these mocks to test code (e.g. `mocks_test.go`)?

chained/water_downloader.go
```diff
@@ -0,0 +1,115 @@
...
+	joinedErrs := errors.New("failed to download WASM from all URLs")
+	for _, url := range d.urls {
+		tempBuffer := &bytes.Buffer{}
+		err := d.downloadWASM(ctx, tempBuffer, url)
+		if err != nil {
+			joinedErrs = errors.Join(joinedErrs, err)
+			continue
+		}
+
+		err = d.verifyHashSum(tempBuffer.Bytes())
```
hwh33: We shouldn't need to implement a checksum - we don't need an independent reliability layer here.  HTTPs or whatever other transports we use should be reliable on their own.

chained/water_downloader.go
```diff
@@ -0,0 +1,115 @@
...
+
+// DownloadWASM downloads the WASM file from the given URLs, verifies the hash
+// sum and writes the file to the given writer.
+func (d *downloader) DownloadWASM(ctx context.Context, w io.Writer) error {
+	joinedErrs := errors.New("failed to download WASM from all URLs")
+	for _, url := range d.urls {
+		tempBuffer := &bytes.Buffer{}
+		err := d.downloadWASM(ctx, tempBuffer, url)
+		if err != nil {
+			joinedErrs = errors.Join(joinedErrs, err)
```
hwh33, line 53: TIL about `errors.Join`!

chained/water_downloader.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func WithExpectedHashsum(hashsum string) DownloaderOption {
+	return func(d *downloader) {
+		d.expectedHashSum = hashsum
+	}
+}
+
+func WithHTTPDownloader(httpDownloader WASMDownloader) DownloaderOption {
```
hwh33: I feel like this probably doesn't need to be included.  What is the use case for this?

chained/proxy.go
```diff
@@ -212,7 +213,7 @@ func createImpl(configDir, name, addr, transport string, s *config.ProxyConfig,
 	case "algeneva":
 		impl, err = newAlgenevaImpl(addr, s, reportDialCore)
 	case "water":
-		impl, err = newWaterImpl(addr, s, reportDialCore)
+		impl, err = newWaterImpl(addr, s, reportDialCore, proxied.DirectThenFrontedClient(1*time.Minute))
```
hwh33: This is a bit nitpicky, but I think we should create the HTTP client in `newWaterImpl`, rather than pass it in.  This will match how other transports do similar things.  Also we're not planning to call `newWaterImpl` with different types of HTTP clients, so there is no need for it to be a parameter.

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+func NewMagnetDownloader(magnetURL string) WASMDownloader {
+	return &magnetDownloader{
+		magnetURL: magnetURL,
+	}
+}
+
+// DownloadWASM downloads the WASM file from the given URL.
+func (d *magnetDownloader) DownloadWASM(ctx context.Context, w io.Writer) error {
+	cfg := torrent.NewDefaultClientConfig()
+	path, err := os.MkdirTemp("/tmp", "torrent")
```
hwh33: From the documentation:
> If dir is the empty string, MkdirTemp uses the default directory for temporary files

Let's do that.  Let's also name the temporary directory something more descriptive like "lantern-water-module".

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+		return fmt.Errorf("failed to create temp dir: %w", err)
+	}
+	cfg.DataDir = path
+	client, err := torrent.NewClient(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create torrent client: %w", err)
+	}
+	defer client.Close()
+
+	t, err := client.AddMagnet(d.magnetURL)
```
hwh33: Suppose we have multiple magnet links for this WASM.  The way the code is currently written, we will create a separate Torrent client for each link.  I think instead we probably want one client configured to use all of the links.

Does that sound correct to you @anacrolix?

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to add magnet: %w", err)
+	}
+	<-t.GotInfo()
```
hwh33: This should be:
```
select {
case <-t.GotInfo():
case <-ctx.Done():
    return fmt.Errorf("context complete: %w", ctx.Error())
}
```

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+	t.DownloadAll()
+	client.WaitAll()
+
+	fmt.Println(t.InfoHash().String())
```
hwh33: Is this debugging code?

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	defer client.Close()
+
+	t, err := client.AddMagnet(d.magnetURL)
+	if err != nil {
+		return fmt.Errorf("failed to add magnet: %w", err)
+	}
+	<-t.GotInfo()
+	t.DownloadAll()
+	client.WaitAll()
```
hwh33: This looks like it will block potentially indefinitely and therefore not honor the provided context.  However, it looks like you can configure an `HTTPDialContext` on the client.  I'm unclear on whether that will apply to this call.  There is also a [`TrackerDialContext`](https://pkg.go.dev/github.com/anacrolix/torrent#ClientTrackerConfig).

hwh33: If those context fields don't apply, there is also [CancelPieces](https://pkg.go.dev/github.com/anacrolix/torrent#Torrent.CancelPieces).  You could call `t.DownloadAll()` and `client.WaitAll` in a separate goroutine, then in the main routine have something like:
```
select {
case <-downloadComplete:
case <-ctx.Done():
    t.CancelPieces(0, t.NumPieces())
}
```

chained/water_magnet_downloader_test.go
```diff
@@ -0,0 +1,43 @@
...
+		assert         func(t *testing.T, r io.Reader, err error)
+	}{
+		{
+			name:           "downloading learning python in a week",
+			givenMagnetURL: "magnet:?xt=urn:btih:34F2A1FA5CD593C394C6E5B5B83B92A7165EA9A9&dn=Learn+Python+In+A+Week+And+Master+It&tr=http%3A%2F%2Fp4p.arenabg.com%3A1337%2Fannounce&tr=udp%3A%2F%2F47.ip-51-68-199.eu%3A6969%2Fannounce&tr=udp%3A%2F%2F9.rarbg.me%3A2780%2Fannounce&tr=udp%3A%2F%2F9.rarbg.to%3A2710%2Fannounce&tr=udp%3A%2F%2F9.rarbg.to%3A2730%2Fannounce&tr=udp%3A%2F%2F9.rarbg.to%3A2920%2Fannounce&tr=udp%3A%2F%2Fopen.stealth.si%3A80%2Fannounce&tr=udp%3A%2F%2Fopentracker.i2p.rocks%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.cyberia.is%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.dler.org%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.internetwarriors.net%3A1337%2Fannounce&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.openbittorrent.com%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=udp%3A%2F%2Ftracker.pirateparty.gr%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.tiny-vps.com%3A6969%2Fannounce&tr=udp%3A%2F%2Ftracker.torrent.eu.org%3A451%2Fannounce",
```
hwh33: Is there any way to do this without relying on an external network resource?  Could we host a torrent locally?  I'm not sure how much work that is.

chained/water_downloader.go
WendelHime: Just for test cases, I need to be able to mock it so the unit test don't become a integration test

chained/proxy.go
WendelHime: I've only added as a parameter because this is the only way I found to make it testable. Hmmmmmm I'll try to create a http client var and check if that works

chained/water_downloader.go
WendelHime: And also because I want to test specifically the downloader logic without touching any implementations

chained/water_magnet_downloader.go
WendelHime: hmmmmm right now, it uses the current directory, that's why I've added to `/tmp` but now I'm wondering about windows

chained/water_magnet_downloader.go
WendelHime: > This looks like it will block potentially indefinitely and therefore not honor the provided context. However, it looks like you can configure an `HTTPDialContext` on the client. I'm unclear on whether that will apply to this call. There is also a [`TrackerDialContext`](https://pkg.go.dev/github.com/anacrolix/torrent#ClientTrackerConfig).

Wasn't aware about that, it should make it easier to test too, thanks!

chained/water_downloader.go
anacrolix: Wait until you discover `fmt.Errorf` now supports multiple `%w`! :)

chained/water_magnet_downloader.go
anacrolix: Yeah it is more efficient. It will sit idle, so leave the client in `magnetDownloader`. That also means it will actually seed stuff for longer than it takes to download too, which I assume is the intention, but fine if not.

chained/water_magnet_downloader.go
anacrolix: Yes, definitely

chained/water_magnet_downloader.go
anacrolix: You might want to use NewReader on the Torrent, and just stream it straight out to the writer `w` here. There is no provided event to wait for completed downloads in anacrolix/torrent, although it's a commonly requested feature. I prefer to point people toward the streaming reader support.

chained/water_magnet_downloader.go
```diff
@@ -0,0 +1,52 @@
...
+
+	fmt.Println(t.InfoHash().String())
+
+	_, err = io.Copy(w, t.NewReader())
```
anacrolix, line 110: Ah, beautiful. This will stream out, you don't need any DownloadAll or anything else, although it doesn't hurt.

go.mod
```diff
@@ -135,22 +137,22 @@ require (
...
+	github.com/anacrolix/missinggo/v2 v2.7.4 // indirect
 	github.com/anacrolix/mmsg v1.0.0 // indirect
 	github.com/anacrolix/multiless v0.3.1-0.20221221005021-2d12701f83f7 // indirect
-	github.com/anacrolix/squirrel v0.6.3 // indirect
+	github.com/anacrolix/squirrel v0.6.4 // indirect
 	github.com/anacrolix/stm v0.4.1-0.20221221005312-96d17df0e496 // indirect
 	github.com/anacrolix/sync v0.5.1 // indirect
-	github.com/anacrolix/torrent v1.53.3 // indirect
-	github.com/anacrolix/upnp v0.1.3-0.20220123035249-922794e51c96 // indirect
+	github.com/anacrolix/torrent v1.53.3
```
anacrolix, line 162: This should move up the block above if you run `go mod tidy`.

anacrolix: It's probably worth updating it too.

WendelHime: I've downgraded to this version because there's an incompatibility with some other service that flashlight depends, there's a commit where I've added a message about it, but I've forgotten to ping you about it 😅

WendelHime: anacrolix/confluence needs a new tag and we need to update that at lantern/replica and then here at lantern/flashlight

WendelHime: Using the latest torrent version breaks the replica dependency

anacrolix: Shieeeet.

chained/water_downloader.go
hwh33: Could it just go in the test code then?

chained/water_magnet_downloader.go
hwh33: > hmmmmm right now, it uses the current directory

That's odd... What do you get if you print `os.TempDir()`?  Do you have $TMPDIR set to "." or something?  You should get the system temporary directory:

> On Unix systems, it returns $TMPDIR if non-empty, else /tmp [[source](https://pkg.go.dev/os#TempDir)]

hwh33:
1. There's a tricky issue here which we'll need to figure out.  Currently, [the client will wait until all dialers have been initialized before proceeding to use them](https://github.com/getlantern/flashlight/blob/292ec2c84c4226fff62fa4238c1f8e67b2244a74/chained/proxy.go#L120).  We'll need to refactor things so that the client can proceed with dialers as soon as they're ready.  I know that adds significant effort to this work, but I don't think the user experience will be reasonable as is.  With the code as it currently stands, the client could stall until the WASM is fully downloaded (or until the download times out).

2. We should also store the downloaded WASMs on disk.  That way we don't have to keep re-downloading the same ones every time the client starts up.  We'll want some kind of versioning and clean-up logic so that we don't hold on to outdated or unused WASMs.  To be more concrete, if we download `water-transport-foo-v2`, then we should delete `water-transport-foo-v1`.  Additionally, if we haven't been assigned a proxy for `water-transport-foobar-v1` in a week, we should delete the WASM.  So I guess to facilitate this, each WASM will need to be uniquely identified (just a name and a version should suffice).

chained/water_magnet_downloader_test.go
WendelHime: I've added some mocks and removed this test

hwh33:
@WendelHime and I discussed the two points above on a call this morning.  We are going to keep the scope of this PR to the current changes and address each of the two points in separate PRs.  We do not have WATER deployed in production, so it will not be a problem to have this code merged in before addressing these two issues.

WendelHime:
Thank you guys for the suggestions! Changes were applied and this is ready for review again!

proxied/proxied.go
```diff
@@ -600,6 +600,19 @@ func DirectThenFrontedClient(timeout time.Duration) *http.Client {
 	}
 }
 
+func ChainedThenDirectThenFrontedClient(timeout time.Duration, rootCA string) *http.Client {
```
hwh33, line 606: Heh, I like it 😄

chained/water_impl.go
```diff
@@ -20,11 +25,36 @@ type waterImpl struct {
...
+		}
+	}
+
+	wasmAvailableAt := ptSetting(pc, "water_available_at")
+	if wasm == nil && wasmAvailableAt != "" {
+		urls := strings.Split(wasmAvailableAt, ",")
+		b := new(bytes.Buffer)
+		cli := httpClient
+		if cli == nil {
+			cli = proxied.ChainedThenDirectThenFrontedClient(1*time.Minute, "")
```
hwh33: These downloads may often take more than 1 minute.  Let's make this part of the proxy configuration or the global configuration.  I'll let you decide which, I think there are pros and cons to both.

proxied/proxied.go
```diff
@@ -600,6 +600,19 @@ func DirectThenFrontedClient(timeout time.Duration) *http.Client {
...
+	drt := &http.Transport{
+		TLSHandshakeTimeout:   timeout,
+		ResponseHeaderTimeout: timeout,
```
hwh33: We could actually hard-code this to something low, like 10 or 30 seconds.  That would allow us to move on to domain fronting faster if we're not able to hit the URL directly.

hwh33 (commented):
Looking good, just a few comments.

proxied/proxied.go
WendelHime: I've hardcoded some timeouts but the client timeout I'll follow the recommendation for using some of the configs.

chained/water_impl.go
WendelHime: I'll choose to add an attribute to lantern-cloud config because this is a specific configuration related to the WATER implementation and we can customize it per protocol/WASM size

hwh33 (approved):
Awesome, LGTM!

anacrolix:
> @WendelHime and I discussed the two points above on a call this morning. We are going to keep the scope of this PR to the current changes and address each of the two points in separate PRs. We do not have WATER deployed in production, so it will not be a problem to have this code merged in before addressing these two issues.

This is absolutely the best way to do PRs. I only have 2 thumbs up to give :D

PR #1426, opened 2024-10-17
Skip failing dialers if we have succeeding ones

myleshorton:
@hwh33 and I observed in user logs how the bandit dialer will continue to retry/explore dialers that are failing to connect at all. 10% of the time, it will attempt dialers that are not the best dialer. Some percentage of those dialers could be not connecting at all. This change ignores those dialers if we have others that are succeeding.

### DIFF
--- a/bandit/bandit.go
+++ b/bandit/bandit.go
@@ -199,9 +199,10 @@ func (o *BanditDialer) chooseDialerForDomain(network, addr string) (Dialer, int)
 	// Loop through the number of dialers we have and select the one that is best
 	// for the given domain.
 	chosenArm := o.bandit.SelectArm(rand.Float64())
+	var dialer Dialer
 	notAllFailing := hasNotFailing(o.dialers)
 	for i := 0; i < (len(o.dialers) * 2); i++ {
-		dialer := o.dialers[chosenArm]
+		dialer = o.dialers[chosenArm]
 		if (dialer.ConsecFailures() > 0 && notAllFailing) || !dialer.SupportsAddr(network, addr) {
 			// If the chosen dialer has consecutive failures and there are other
 			// dialers that are succeeding, we should choose a different dialer.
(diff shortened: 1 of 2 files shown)


myleshorton:
hmnn...test fails some of the time...investigating

bandit/bandit.go
```diff
@@ -178,21 +178,42 @@ func (o *BanditDialer) DialContext(ctx context.Context, network, addr string) (n
...
+		if (dialer.ConsecFailures() > 0 && notAllFailing) || !dialer.SupportsAddr(network, addr) {
+			// If the chosen dialer has consecutive failures and there are other
+			// dialers that are succeeding, we should choose a different dialer.
+			//
+			// If the chosen dialer does not support the address, we should also
+			// choose a different dialer.
+			chosenArm = differentArm(chosenArm, len(o.dialers))
+		} else {
+			break
+		}
```
WendelHime: This suggestion is totally optional since the code already works as intended
```suggestion
                         continue
		}
	        break
```

bandit/bandit_test.go
```diff
@@ -14,11 +14,80 @@ import (
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			opts := Options{
+				Dialers: tt.fields.dialers,
+			}
+			o, err := New(opts)
+			if err != nil {
+				t.Fatal(err)
+			}
```
WendelHime:
```suggestion
			require.NoError(t, err)
```

bandit/bandit_test.go
```diff
@@ -14,11 +14,80 @@ import (
...
+			got, got1 := o.chooseDialerForDomain(tt.args.network, tt.args.addr)
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("BanditDialer.chooseDialerForDomain() got = %v, want %v", got, tt.want)
+			}
+			if got1 != tt.want1 {
+				t.Errorf("BanditDialer.chooseDialerForDomain() got1 = %v, want %v", got1, tt.want1)
+			}
```
WendelHime: Please don't commit this suggestion directly from github otherwise it'll break/panic since we are not importing `"github.com/stretchr/testify/assert"`  at this file yet
```suggestion
                        assert.Equal(t, tt.want1, got1, "BanditDialer.chooseDialerForDomain() got1 = %v, want %v", got1, tt.want1)
```

WendelHime (approved):
Added a few optional suggestion but LGTM!

myleshorton:
OK thanks @WendelHime I added those changes! Going to test a little bit more then bring this in.

myleshorton:
OK this is working well for me.

PR #1427, opened 2024-10-17
adding water version control

WendelHime:
This PR requires #1425 to be merged before merging this. This pull requests implements a version control for WASM files, the goal is to store WASM files locally and manage their versions.

- When a protocol is loaded correctly it should write at the history CSV and mark which WASM file was loaded so it can use it as reference for deleting later.
- When a protocol is not being used/loaded in the last week, it should be deleted.
- If the WASM file is not available locally, it should use the downloader for retrieving the newest version.


chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+	"slices"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/getlantern/flashlight/v7/proxied"
+	"github.com/gocarina/gocsv"
+)
+
+type versionControl struct {
```
hwh33: Since this is specific to WATER and in a package with a bunch of non-WATER stuff, could you name it to reflect this?  Probably `waterVersionControl`?

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+type wasmInfo struct {
+	version   string
+	protocol  string
+	builtWith string
+	path      string
+}
+
+type VersionControl interface {
```
hwh33: There is only one implementation of this interface, so let's do away with it and just use the implementation directly.

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+	defer f.Close()
+
+	if err := gocsv.MarshalFile(&vc.history, f); err != nil {
+		return log.Errorf("failed to marshal file %s: %w", historyCSV, err)
+	}
+
+	return nil
+}
+
+func (vc *versionControl) downloadWASM(ctx context.Context, transport string, urls []string) (wasmInfo, error) {
```
hwh33: What if you gave it a `WASMDownloader`?  Then `versionControl` can just focus on version control stuff and doesn't need to worry about how the download actually works (e.g. the chained-then-direct-then-fronted stuff).

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+type VersionControl interface {
+	GetWASM(ctx context.Context, transport string, urls []string) (io.ReadCloser, error)
+	Commit(transport string) error
+}
+
+// NewVersionControl check if the received config is the same as we already
+// have and if not, it'll try to fetch the newest WASM available.
+func NewVersionControl(configDir string) (VersionControl, error) {
```
hwh33: This isn't referenced outside of this package, so let's not export it.  The same is true for a number of WATER related constructs actually, like WASMDownloader (which I realize was introduced before this PR; I didn't catch it).  Could you change all of those to be unexported please?

chained/water_impl.go
```diff
@@ -39,40 +38,44 @@ func newWaterImpl(addr string, pc *config.ProxyConfig, reportDialCore reportDial
...
-				var err error
-				timeout, err = time.ParseDuration(downloadTimeout)
-				if err != nil {
-					return nil, fmt.Errorf("failed to parse download timeout: %w", err)
-				}
-			}
-
-			cli = proxied.ChainedThenDirectThenFrontedClient(timeout, "")
+		waterDir := filepath.Join(configDir, "water")
+		vc, err := NewVersionControl(waterDir)
```
hwh33: Could you instead have `newWaterImpl` just accept a path to a version control directory?  `newWaterImpl` doesn't need to know about the existence of a config directory or manage anything in the directory containing the version control folder.

chained/water_impl.go
```diff
@@ -39,40 +38,44 @@ func newWaterImpl(addr string, pc *config.ProxyConfig, reportDialCore reportDial
...
+		if err != nil {
+			return nil, log.Errorf("failed to read wasm: %w", err)
 		}
-		wasm = b.Bytes()
+
+		if len(b) == 0 {
+			return nil, log.Errorf("received empty wasm")
+		}
+
+		if err = vc.Commit(transport); err != nil {
```
hwh33: Couldn't `GetWASM` do this?  Any time a WASM is fetched, we know it's being used.

chained/water_version_control.go
hwh33: Also the input directory is not the config directory, nor should the WATER version control system need to know anything about the config directory.  You could just call it `dir`.

chained/water_version_control.go
hwh33: Likewise, `GetWASM` (which calls this function) could accept a `WASMDownloader`.

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+
+type history struct {
+	Transport      string    `csv:"transport"`
+	LastTimeLoaded time.Time `csv:"last_time_loaded"`
+}
+
+type wasmInfo struct {
+	version   string
+	protocol  string
+	builtWith string
```
hwh33: It doesn't look like these fields are actually used anywhere.  All we're actually doing is naming the files based on whatever the back-end tells us (and insisting these names contain two periods), then storing the WASM in those files.  Let's just keep it simple and not try to interpret the names at all.

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+)
+
+type versionControl struct {
+	// contain all wasm files available locally
+	// key is something like: protocol.version
+	wasmFilesAvailable map[string]wasmInfo
+	dir                string
+
+	// history of the last time the WASM file was loaded
+	history      []history
```
hwh33: You're already maintaining a map of transport names to info about the transport.  Why not just track this information there?  This would simplify the logic quite a bit.  Combining this with a simplification of `wasmInfo` (suggested below), you'd have:

```
type versionControl struct {
    dir string
    wasms map[string]wasmInfo
}

type wasmInfo struct {
    lastLoaded time.Time
    path string
}
```

If you'd like, you could simplify it even further by just relying on the directory structure.  If your directory looked like:
```
dir
    ├── transport-foo.wasm
    ├── transport-foo.last-loaded
    ├── transport-bar.wasm
    └── transport-bar.last-loaded
```

Then your version control system could just be:
```
type versionControl string

// Essentially just documenting that versionControl represents the directory.
func newVersionControl(dir string) versionControl {
    return versionControl(dir)
}

func (vc versionControl) getWASM(ctx context.Context, transport string, downloader wasmDownloader) (io.ReadCloser, error) {
    path := filepath.Join(vc, transport + ".wasm")
    f, err := os.Open(path)
    if errors.Is(err, fs.ErrNotExist) {
        f, err = os.Create(path)
        // check and handle err
        err = downloader.downloadWASM(ctx, f)
        // check and handle err
        _ , err = f.Seek(0, 0)
        // check and handle err
    }
    vc.markUsed(transport)
    return f, nil
}

func (vc versionControl) markUsed(transport string) {
    f, err := os.Create(filepath.Join(vc, transport + ".last-loaded"))
    // check and log err
    // write time.Now() to f
    go vc.cleanOutdated()
}

func (vc versionControl) cleanOutdated() {
    // Walk the directory, parsing ".last-loaded" files and deleting ".wasm" files as necessary.
}
```

chained/water_version_control.go
```diff
@@ -0,0 +1,274 @@
...
+
+	vc.history = history
+	return nil
+}
+
+func (vc *versionControl) deleteOutdatedWASMFiles() error {
+	affectedTransports := make([]string, 0)
+	for _, h := range vc.history {
+		// check if the file is outdated
+		sevenDaysAgo := time.Now().AddDate(0, 0, -7)
```
hwh33: Could you move this constant to the top of the file and make it clear what it's used for?  Something like:
```
const unusedWASMsDeletedAfter = 7 * 24 * time.Hour
```

chained/water_impl.go
WendelHime: In this case, how should we set this? Which package or function should be responsible for setting the water dir? `chained/proxies.go` refers to `newWaterImpl` for initializing, should it be responsible for setting the waterDir?

chained/water_version_control.go
```diff
@@ -0,0 +1,136 @@
...
+// Remember to Close the io.ReadCloser after using it
+func (vc *waterVersionControl) GetWASM(ctx context.Context, transport string, downloader waterWASMDownloader) (io.ReadCloser, error) {
+	path := filepath.Join(vc.dir, transport+".wasm")
+	var f io.ReadCloser
+	f, err := os.Open(path)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, log.Errorf("failed to open file %s: %w", path, err)
+	}
+
+	if os.IsNotExist(err) {
```
hwh33: From the godoc:

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

That said, I don't imagine this will create real problems here.  Just an FYI really.

chained/water_version_control.go
```diff
@@ -0,0 +1,136 @@
...
+				transport := strings.TrimSuffix(filepath.Base(path), ".last-loaded")
+				if err = os.Remove(filepath.Join(vc.dir, transport+".wasm")); err != nil {
+					log.Errorf("failed to remove wasm file %s: %w", transport+".wasm", err)
+					return
+				}
+				if err = os.Remove(path); err != nil {
+					log.Errorf("failed to remove last-loaded file %s: %w", path, err)
+					return
+				}
+			}()
```
hwh33: I wonder if this could create issues with the ongoing `filepath.Walk` loop.  You're concurrently deleting files in the same directory you're walking.  I have no idea really.  I guess it depends on what's going on under the hood in `filepath.Walk`.  Might be safer though to do all of the deleting after the `filepath.Walk` call exits.

chained/proxy.go
```diff
@@ -212,7 +212,7 @@ func createImpl(configDir, name, addr, transport string, s *config.ProxyConfig,
 	case "algeneva":
 		impl, err = newAlgenevaImpl(addr, s, reportDialCore)
 	case "water":
-		impl, err = newWaterImpl(addr, s, reportDialCore)
+		impl, err = newWaterImpl(configDir, addr, s, reportDialCore)
```
hwh33: Looks like this will now populate the config directory with .wasm and .last-loaded files.  I think we still want a sub-directory for those files.  You asked [here](https://github.com/getlantern/flashlight/pull/1427#discussion_r1815590387) where this directory should be specified and created.  I think this is an appropriate location.  How about something like:
```
waterDir := filepath.Join(configDir, "water")
if err := os.Mkdir(waterDir, 0755); err != nil && !errors.Is(err, fs.ErrExist) {
    return nil, fmt.Errorf("failed to create water directory: %w", err)
}
impl, err = newWaterImpl(waterDir, addr, s, reportDialCore)
```

hwh33 (commented):
Just a couple of small things!

chained/water_impl.go
hwh33: Answered in another comment [here](https://github.com/getlantern/flashlight/pull/1427/files#r1817085389).

chained/water_version_control.go
WendelHime: I'll follow the docs recommendations! Thanks for reminding!

chained/proxy.go
garmr-ulfr: Just FYI `err != nil` isn't necessary, `errors.Is` checks for `nil`.

garmr-ulfr: Also, `os.MkdirAll` won't return an error if the directory already exists. So you could drop `err != nil` if using `os.Mkdir` or `!errors.Is(err, fs.ErrExist)` if you use `os.MkdirAll`. This is just an FYI!

WendelHime: Moving to `os.MkdirAll`, thanks for the suggestion!

hwh33: > Just FYI err != nil isn't necessary, errors.Is checks for nil.

The check is for the negation.  `!errors.Is(err, fs.ErrExist)` on its own will miss the case in which we have a non-nil error due to something other than `fs.ErrExist`.

You are correct though that `err != nil && errors.Is(err, SomeError)` is equivalent to `errors.Is(err, SomeError)`.

> Also, os.MkdirAll won't return an error if the directory already exists.

It will.  Try the following:

```
for i := 0; i < 2; i++ {
	if err := os.Mkdir("test-dir", 0755); err != nil {
		fmt.Println(err)
		fmt.Println(errors.Is(err, fs.ErrExist))
	}
}
```

You'll get:

```
mkdir test-dir: file exists
true

File: chained/proxy.go [hunk 13]
```diff
@@ -212,9 +214,13 @@ func createImpl(configDir, name, addr, transport string, s *config.ProxyConfig,
 	case "algeneva":
 		impl, err = newAlgenevaImpl(addr, s, reportDialCore)
 	case "water":
-		impl, err = newWaterImpl(addr, s, reportDialCore)
+		waterDir := filepath.Join(configDir, "water")
+		if err := os.MkdirAll(waterDir, 0755); err != nil {
```
hwh33, line 218: This will fail when the directory already exists.  Explanation [here](https://github.com/getlantern/flashlight/pull/1427#discussion_r1817239347).

chained/proxy.go
hwh33: Oh maybe you're thinking of `os.MkdirAll`.  That won't error if the directory already exists.  You should use that @WendelHime.  Then you can use @garmr-ulfr's suggested simplification.

hwh33: Wait I'm a dummy.  `os.MkdirAll` is what you suggested @garmr-ulfr.

Reading comprehension... Sorry for the confusion 😬 🤦

chained/proxy.go
hwh33: Sorry, ignore me, I was mixed up.  This looks good!

hwh33 (approved):
LGTM!

chained/proxy.go
garmr-ulfr: > > Just FYI err != nil isn't necessary, errors.Is checks for nil.
> 
> The check is for the negation. `!errors.Is(err, fs.ErrExist)` on its own will miss the case in which we have a non-nil error due to something other than `fs.ErrExist`.

You're correct. `errors.Is(err, fs.ErrExist)` would return false because `nil != fs.ErrExist`. Ignore that part of what I said. 

> Wait I'm a dummy. `os.MkdirAll` is what you suggested @garmr-ulfr.
> 
> Reading comprehension... Sorry for the confusion 😬 🤦

:laughing: I had my own "dummy" moment. We'll just call it even. It's Friday.

PR #1429, opened 2024-10-18
loading water dialer async

WendelHime:
This pull request requires #1425 and #1427 to be merged.

This pull request goal is to load the WATER dialer asynchronously so when flashlight start initializing all proxies that it could find it doesn't keep waiting until WATER dialer is ready (it can take some time since the WASM need to be downloaded)


chained/water_impl.go
```diff
@@ -42,54 +44,86 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
+	return dialer, nil
 }
 
 func (d *waterImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
 	return d.reportDialCore(op, func() (net.Conn, error) {
+		// TODO check if this is the intended behavior or if we should just return an error
+		d.wgDownload.Wait()
+		if d.dialer == nil {
+			return nil, log.Errorf("dialer not available")
+		}
```
garmr-ulfr: Check if what is the intended behavior? There are places where `d.dialer` can be `nil`, so unless I'm missing something, this is correct. Although, we should wrap the error that occurred while trying to download the WASM. That way, we would have a bit more info on why it failed in the traces instead of only in the logs. 
Maybe something like:
```
d.wgDownload.Wait()
if d.dialer == nil {
	d.downloadErrMx.Lock()
	err := d.downloadErr
	d.downloadErrMx.Unlock()

	return nil, log.Errorf("dialer not available: %w", err)
}
```

chained/water_impl.go
```diff
@@ -42,54 +44,86 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
+			return nil, log.Errorf("failed to create dialer: %w", err)
 		}
+		d.dialer = dialer
+	}
 
-		wasm = b
+	if wasm == nil && wasmAvailableAt != "" {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
garmr-ulfr, line 63: Nit: Why a `WaitGroup`? Since it's only one goroutine we'd be waiting on, we should probably use a `chan struct{}` to signal it's done. I know it's a small thing but, it would help with readability and maintainability so the reader knows it's only one goroutine.

chained/water_impl.go
```diff
@@ -42,54 +44,86 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
-
-		if len(b) == 0 {
-			return nil, log.Errorf("received empty wasm")
+			return nil, log.Errorf("failed to create dialer: %w", err)
 		}
+		d.dialer = dialer
+	}
 
-		wasm = b
+	if wasm == nil && wasmAvailableAt != "" {
```
garmr-ulfr: Nit: Isn't this basically `else if wasmAvailableAt != ""`?

garmr-ulfr (commented):
Just a couple of things.

chained/water_impl.go
WendelHime: I've chosen to use a wait group because I can use it at tests to know when it loaded ~~correctly~~ the WASM file (when downloading). But it also makes sense to use a channel

chained/water_impl.go
WendelHime: I'll follow the suggestion for adding the error, but I don't believe we need the mutex in this case, since the waitgroup is already saying the operation is done, the error wouldn't be accessed or modified anywhere else.

garmr-ulfr: > I don't believe we need the mutex in this case, since the waitgroup is already saying the operation is done

Oh, right. That's a good point.

chained/water_impl.go
garmr-ulfr: Oh OK, that makes sense. This is fine as is and it works, so it's not an issue!

garmr-ulfr (approved):
LGTM

chained/water_impl.go
```diff
@@ -22,74 +23,105 @@ type waterImpl struct {
...
-		raddr:          addr,
-		dialer:         dialer,
-		reportDialCore: reportDialCore,
-	}, nil
+	return dialer, nil
 }
 
 func (d *waterImpl) dialServer(op *ops.Op, ctx context.Context) (net.Conn, error) {
 	return d.reportDialCore(op, func() (net.Conn, error) {
+		d.wgDownload.Wait()
```
hwh33, line 120: Hey @WendelHime sorry I just saw this logic.  Won't this lead to the client still getting blocked until the WASM download finishes?  The client will try to use this dialer, but the call to `waterImpl.dialServer` will block on this line.  I think you need to respect the input context, like:
```
select {
case <-someChannelWaitingOnWaterDownload:
    // carry on with dial
case <-ctx.Done():
    return nil, log.Errorf("context completed while waiting for WASM download: %w", ctx.Error())
}
```

However, even that has issues as the WATER dialer will then appear to be non-functional.  In the current code, the bandit dialer will simply [ignore it](https://github.com/getlantern/flashlight/blob/7fbdf0b33ce9bf0a8d13a7c66a90ab1d06b0b01d/bandit/bandit.go#L208-L213).  Even if that code changes, errors returned by a proxy dialer are always going to lead to that dialer being de-prioritized.  We don't want the WATER dialer to always get ignored or de-prioritized.

Instead, we need the WATER dialer to be added to the overall set of dialers only when it's ready for use.  I realize that is a more involved change.  Sorry that this comment is coming after this work was already reviewed and merged.

WendelHime: 🤔  can we introduce a new dialer function flag that says the dialer is ready? I'm considering these steps:

1. Create a `IsReady() bool` function at [bandit dialer interface](https://github.com/getlantern/flashlight/blob/main/bandit/bandit.go#L275).
2. Implement `IsReady` to all chained implementations.
3. Update water impl dialer to update the status properly.
4. Update [bandit logic for checking](https://github.com/getlantern/flashlight/blob/main/bandit/bandit.go#L207), if the dialer is not ready continue search of dialers.

This solve the issue for bandit, but not for bypass

hwh33: I like that approach!  Looks like bypass will get access to that function as well since `chained.CreateDialersMap` returns a map of `bandit.Dialers`.

garmr-ulfr: > In the current code, the bandit dialer will simply [ignore it](https://github.com/getlantern/flashlight/blob/7fbdf0b33ce9bf0a8d13a7c66a90ab1d06b0b01d/bandit/bandit.go#L208-L213). Even if that code changes, errors returned by a proxy dialer are always going to lead to that dialer being de-prioritized. We don't want the WATER dialer to always get ignored or de-prioritized.

Sorry, this had not occurred to me. Good catch!

WendelHime: PR ready for review at: https://github.com/getlantern/flashlight/pull/1436

PR #1436, opened 2024-11-06
adding IsReady to dialer interface

WendelHime:
This pull request aim to fix the issues brought by this [comment](https://github.com/getlantern/flashlight/pull/1429/files#r1829658328). Now we should be able to ignore dialers when they're not ready to be called


bypass/bypass.go
```diff
@@ -84,24 +84,43 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
+			log.Debugf("dialer %q is not ready, starting in background", name)
+			go func() {
+				for {
+					time.Sleep(60 * time.Second)
+					if dialer.IsReady() {
+						b.startProxy(name, config, configDir, userConfig, dialer)
+						break
+					}
+				}
+			}()
```
garmr-ulfr: We should probably have a way to stop this in the event a `dialer` never becomes ready or becomes ready after new proxies are already assigned. It's kind of unlikely a `dialer` is never ready, but if someone (probably me) forgets/unaware this goroutine is created, they might not add a timeout to the `dialer` themselves. 

Maybe a `chan` or `ctx` that gets closed/canceled in `Reset` and is recreated if we need to wait for `dialer`s again?

chained/water_impl_test.go
```diff
@@ -69,6 +69,10 @@ func TestNewWaterImpl(t *testing.T) {
 			assert: func(t *testing.T, actual *waterImpl, err error) {
 				require.NoError(t, err)
 				require.NotNil(t, actual)
+				select {
+				case <-actual.dialerAvailable:
+					// waiting download to be finished
+				}
```
garmr-ulfr: This could be simplified to just 
```
<-actual.dialerAvailable
```
no need for the select/case.

chained/water_impl_test.go
```diff
@@ -101,10 +105,14 @@ func TestNewWaterImpl(t *testing.T) {
...
+				defer actual.close()
 				require.NoError(t, err)
 				require.NotNil(t, actual)
-				actual.wgDownload.Wait()
-				assert.NoError(t, actual.downloadErr)
+				assert.NoError(t, actual.errLoadingWASM)
+				select {
+				case <-actual.dialerAvailable:
+					// waiting download to be finished
+				}
```
garmr-ulfr: Same as above.

chained/water_impl.go
```diff
@@ -36,31 +37,41 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
 	transport := ptSetting(pc, "water_transport")
 	wg := new(sync.WaitGroup)
```
garmr-ulfr: This could be removed now since there isn't anything waiting for it.

chained/water_impl.go
```diff
@@ -36,31 +37,41 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
+		}()
 		return d, nil
 	}
 
 	if wasmAvailableAt != "" {
 		wg.Add(1)
 		go func() {
-			defer wg.Done()
+			defer func() {
+				d.dialerAvailable <- struct{}{}
```
garmr-ulfr: I think this is meant to be right before/after `d.ready = true` (line 112) instead of deferring it. This will send over `dialerAvailable` even if there's an error.

garmr-ulfr: Also, this will panic if `close` is called before this sends. We should wrap it in a `select/case`. Same for above.

chained/water_impl.go
```diff
@@ -115,11 +129,31 @@ func createDialer(ctx context.Context, wasm []byte, transport string) (water.Dia
...
 	return d.reportDialCore(op, func() (net.Conn, error) {
-		d.wgDownload.Wait()
-		if d.dialer == nil {
-			return nil, log.Errorf("dialer not available: %w", d.downloadErr)
+		// if dialer is not ready, wait until WASM is downloaded or context timeout
+		if !d.isReady() {
+			select {
+			case <-d.dialerAvailable:
+				log.Debug("download finished!")
+				// carry on with dial
```
garmr-ulfr: You missed the `ok` check here so this doesn't continue to dial if `close` is called while it's waiting.

garmr-ulfr (commented):
This looked like it was tedious having to go add `isReady` to every `dialer`! Good start though. There are a couple of bugs with `dialerAvailable`; I think [getlantern/eventual](https://github.com/getlantern/eventual) would fix them and might even simplify everything a bit. Your way works too! Just an FYI if you weren't already aware of it. It's really useful and deals with a lot of the boilerplate of waiting on values with a timeout.

bypass/bypass.go
WendelHime: Hmmmmm thanks for the suggestion! I'll make IsReady return an error too, so if err is nil and ok is false then it should try again.

WendelHime: In this way we can also customize when to retry to start a proxy or call the func to create a implementation again depending of the returned error

chained/water_impl.go
WendelHime: hmmmm maybe I named wrong, this channel is supposed to notify when it finishes to load the WASM/dialer so it's ok to be returned when there's an error. I'll rename it

bypass/bypass.go
garmr-ulfr: This will still suffer from the same problem though; It's not guaranteed that `isReady` will eventually return an it's ready or an error. If, at some point, we were to update `water` to retry downloading the WASM when it fails, or add a new protocol that might not be ready immediately, the caller can't assume that the dialer itself isn't running an infinite loop. Or there might be a bug where `isReady` never returns true or a non nil error. The goroutines won't ever be cleaned up, leading to memory leaks that we may not catch.

WendelHime: Hmmmm it makes sense, thanks catching this! I've updated the bypass code:

1. Added a go routine function `loadProxyAsync` for creating a context that timeout after 10 min.
2. The function also provides an atomic boolean for the internal go routine that actually check if the proxy is ready or not. So the `loadProxyAsync` waits until the context timeout or it receives a channel signal saying it's ready. When ready the internal go routine should stop looping with the flag marked as false and the ready channel is closed.

WendelHime:
Some changes have been implemented at this PR and fix a data race that was happening between bypass and bandit when initializing the WATER proxy. Both basically start at the same time and it was starting to download and truncating the downloaded data making one of them unusable, so I've added a package-level mutex to make sure it runs one at time.

bandit/bandit.go
```diff
@@ -204,12 +204,22 @@ func (o *BanditDialer) chooseDialerForDomain(network, addr string) (Dialer, int)
...
+		}
+
+		if (dialer.ConsecFailures() > 0 && notAllFailing) || !dialer.SupportsAddr(network, addr) || !ready {
 			// If the chosen dialer has consecutive failures and there are other
 			// dialers that are succeeding, we should choose a different dialer.
 			//
 			// If the chosen dialer does not support the address, we should also
 			// choose a different dialer.
+			//
+			// If the dialer isn't ready we should also choose a another dialer.
```
hwh33: Typo: "choose ~a~ another dialer"

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
-		b.proxies = append(b.proxies, p)
-		go p.start()
+		// if dialer is not ready, try to load it async
+		if ready, err := dialer.IsReady(); err == nil && !ready {
+			log.Debugf("dialer %q is not ready, starting in background", name)
+			go b.loadProxyAsync(name, config, configDir, userConfig, dialer)
+			continue
+		}
+
+		b.startProxy(name, config, configDir, userConfig, dialer)
```
hwh33: It looks like this misses the case where `dialer.IsReady` returns an error.  Is that intentional?

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
+	defer cancel()
+	readyChan := make(chan struct{})
+	retry := atomic.NewBool(true)
+	go func() {
+		for retry.Load() {
+			time.Sleep(15 * time.Second)
+			ready, err := dialer.IsReady()
+			if err != nil {
+				log.Errorf("dialer isn't ready and returned an error: %w", err)
+				break
```
hwh33: Exiting early here without alerting the main thread will cause the main thread to wait the full 10 minutes for the context to time out.

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
 	}
 }
 
+func (b *bypass) loadProxyAsync(proxyName string, config *commonconfig.ProxyConfig, configDir string, userConfig common.UserConfig, dialer bandit.Dialer) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+	readyChan := make(chan struct{})
+	retry := atomic.NewBool(true)
+	go func() {
+		for retry.Load() {
```
hwh33: `ctx.Done` already gives you a way to stay in touch with the main thread.  You could do away with `retry` and add a check of `ctx.Done` to every loop:
```
select {
case <-ctx.Done():
  return
default:
  // carry on
}
```

A few more lines of code, but less variables, less synchronization, and more conventional.  Your call though.

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
+			time.Sleep(15 * time.Second)
+			ready, err := dialer.IsReady()
+			if err != nil {
+				log.Errorf("dialer isn't ready and returned an error: %w", err)
+				break
+			}
+			if !ready {
```
hwh33: Should be `if ready`, right?

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
+				readyChan <- struct{}{}
+				break
+			}
+		}
+	}()
+	select {
+	case _, ok := <-readyChan:
+		if !ok {
+			log.Errorf("ready channel for proxy %q is closed", proxyName)
+		}
```
hwh33: Nobody closes `readyChan` except this thread, so there's no need to check for this here.

bypass/bypass.go
```diff
@@ -84,24 +84,67 @@ func (b *bypass) OnProxies(infos map[string]*commonconfig.ProxyConfig, configDir
...
+	case <-ctx.Done():
+		log.Errorf("proxy %q took to long to get ready", proxyName)
+	}
+	retry.Store(false)
+	close(readyChan)
```
hwh33: This creates a race condition in which the following happens:
- Secondary thread loads `retry`, sees `true` and enters the loop.
- Primary thread sets `retry` to false.
- Primary thread closes `readyChan`.
- Secondary thread tries to send on `readyChan`.
- Panic! 😱 

I would just remove this call to close `readyChan`.  Open channels do not cause any kind of leak; they will be fully cleaned up by the garbage collector.  A couple rules of thumb to help with this in the future:
- Only close a channel if necessary to send a signal to consumers of the channel.
- Only writers (goroutines which are sending on the channel) should close the channel.

I would also remove `retry` as suggested above.  It's not the primary cause of the race condition here, but its presence contributes to the complexity and therefore confusion IMO.

chained/water_impl.go
```diff
@@ -23,85 +23,107 @@ type waterImpl struct {
...
+			d.setReady()
 		}()
 	}
 
 	return d, nil
 }
 
+func (d *waterImpl) loadWASM(ctx context.Context, transport string, dir string, wasmAvailableAt string) (io.ReadCloser, error) {
+	waterLoadingWASMMutex.Lock()
+	defer waterLoadingWASMMutex.Unlock()
```
hwh33: This mutex is just to protect the backing file, right?  I didn't think about that when looking at this work the first time.  Good catch!  That said, locking here like this will force the client to download separate WASMs in serial.  We don't want that...

I think we can use OS synchronization on the files instead of trying to coordinate the synchronization ourselves.  So:
1. As we are already doing [here](https://github.com/getlantern/flashlight/blob/392245a95208c9b46348387ff3118caaa19266b6/chained/water_version_control.go#L37-L47), we will first check to see whether the WASM file already exists.  If not, we will proceed to step 2.
2. When we create the WASM file [here](https://github.com/getlantern/flashlight/blob/392245a95208c9b46348387ff3118caaa19266b6/chained/water_version_control.go#L132), use `os.OpenFile(filename, O_CREATE | O_EXCL, 0644)`.  This will succeed if and only if the file does not yet exist.  If that call fails with `fs.ErrExist`, this means that somebody beat us to the task of writing this file.  When this happens, we go back to step 1.

There's still an edge case here in which somebody opens the file for reading while someone else is still writing it.  They could read an incomplete WASM and initialize the transport incorrectly.  That's tricky...  Maybe we could use a temporary file to indicate when a write is in progress (e.g. `shadowsocks-1.0.0.wasm.in-progress`)?  Something like:
- When we try to read the file:
  - Start with `os.Open(filename)`.
  - If `os.Open(filename)` fails because the file does not exist, proceed to downloading and writing.
  - If `os.Open(filename)` is successful, check for existence of the in-progress file.
  - If in-progress file exists, periodically check on it until it goes away.
  - When in-progress file is gone, we are safe to read.
- When we write the file:
  - We _first_ call `os.OpenFile(inProgressFilename, O_CREATE | O_EXCL, 0644)`.  If successful, we own the in-progress file and are the only ones allowed to read or write the file.  If unsuccessful, somebody beat us to writing, so we go back to reading.
  - If we are successful in creating the in-progress file, we own the WASM file, so we can create it with `os.Create(wasmFilename)` and write to it when we're ready.
  - When we're done writing, we delete the in-progress file to signal that the WASM file is ready for reading.

Does that make sense?  Does that sound correct?  I think that should resolve any concurrency issues without the need for synchronization within the runtime (using mutexes as we do here for example).  As a somewhat helpful optimization, we can have writers establish ownership of the in-progress file before starting to download the WASM.  That will prevent cases in which multiple goroutines redundantly download the same WASM in parallel.

hwh33: Hmm but then there's an edge case in which the application dies before removing the in-progress file or completing the write of the WASM file.  I'm having trouble coming up with a foolproof solution for that.  Maybe in-memory locks are the way to go since we should only ever expect to have one instance of Lantern messing with these files.

Still, we need something more sophisticated than a global lock over the entire WASM download system I think.

chained/water_impl.go
```diff
@@ -23,85 +23,107 @@ type waterImpl struct {
...
+	}
+	return r, nil
+}
+
+func (d *waterImpl) finishedLoading() {
+	select {
+	case d.finishedToLoad <- struct{}{}:
+	default:
+		log.Error("channel closed")
+	}
```
hwh33: There is no safe way to write to a channel if you don't know whether it's closed.  This call should only happen if we are sure the channel is open.  To avoid this dilemma, only the goroutine responsible for writing to a channel should close that channel (if it needs to).

Here's an example of this kind of `select` statement resulting in a panic: https://go.dev/play/p/SUgKWfok-1Y

chained/water_impl.go
```diff
@@ -115,11 +137,45 @@ func createDialer(ctx context.Context, wasm []byte, transport string) (water.Dia
...
+		if err != nil {
+			return nil, log.Errorf("failed to load WASM for dialer: %w", err)
+		}
+
+		if !ready {
+			select {
+			case _, ok := <-d.finishedToLoad:
+				if !ok {
+					d.errLoadingWASM = log.Error("Channel closed")
+					return nil, d.errLoadingWASM
```
hwh33: This error is kind of mysterious to callers of `dialServer` since they don't know anything about a channel or why it would be closed.  The channel should only be closed if `waterImpl` is closed, right?  If that is correct, then maybe this error should read "dialer closed".

hwh33: It also looks like there is unprotected writing of `d.errLoadingWASM` happening here.  `dialServer` could be called concurrently by multiple goroutines, right?

chained/water_version_control.go
```diff
@@ -32,24 +32,48 @@ func newWaterVersionControl(dir string) *waterVersionControl {
...
+
+		return response, nil
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, log.Errorf("failed loading file info")
+	}
+
+	if fi.Size() == 0 {
```
hwh33: How could this happen?  How do we know it's not going to happen again in the second `downloadWASM` call?  This seems like an unnecessary and confusing edge case to handle.  Am I missing something?

bandit/bandit.go
```diff
@@ -353,6 +363,9 @@ type Dialer interface {
 	// connections created via this dialer.
 	DataRecv() uint64
 
+	// IsReady indicates when the dialer is ready for dialing
+	IsReady() (bool, error)
```
hwh33: There's a lot of complexity (in the bandit and bypass code) created by two things:
1. The double return value here.  The first value and second value each need to be checked independently before calling code can proceed to use the dialer.
2. A lack of synchronization provided by this function.  Callers are left having to poll this function.  Since we're trying to communicate when an event happens (when the dialer is ready), this feels like a natural use case for a channel.

Maybe instead, something like:
```
// Ready returns a channel which will have a value on it only when initialization
// of the dialer is complete. If initialization failed, the channel will have a non-nil
// error value.
Ready() chan<- error
```

Then callers use this like:
```
select {
case err := <-dialer.Ready():
    if err != nil {
        return fmt.Errorf("dialer initialization failed: %w", err)
    }
case <-ctx.Done():
    return ctx.Error()
}
// Here we know the dialer is ready to use.
```

Your way definitely works, to be clear!  I'm fine with that if you prefer it.

hwh33 (commented):
I think the synchronization here needs a little work, but I like the overall direction!

bypass/bypass.go
WendelHime: hmmmmm nop, I should check it, thanks for noticing this!

chained/water_impl.go
WendelHime: 🤔 We could implement a locker map for WATER transports. If the app stops before downloading we could check when the file was loaded successfully (or check if the file exists), if it wasn't we could try to download it again.

chained/water_impl.go
WendelHime: > It also looks like there is unprotected writing of d.errLoadingWASM happening here. dialServer could be called concurrently by multiple goroutines, right?

You're right, that validation should only happen when it receives the signal that it finished to load. Thanks for pointing that!

WendelHime: Wait, nop, this verification is only happening after receiving a `ready` dialer. We probably don't need this verification

chained/water_version_control.go
WendelHime: This was happening without the locker/concurrency issues, I should update this so we can handle if the WASM wasn't loaded instead of checking the size

bandit/bandit.go
WendelHime: Hmmmmm this is valid and would work perfectly on the bandit side, but on bypass, how do we know when we should start the proxy asynchronously? Should we start all proxies asynchronously?

WendelHime: hmmmm I'll return nil by default, if the chan is nil It's ready by default

dialer/dialer.go
```diff
@@ -212,6 +212,11 @@ type ProxyDialer interface {
 	// connections created via this dialer.
 	DataRecv() uint64
 
+	// Ready returns a channel which will have a value on it only when initialization
+	// of the dialer is complete. If initialization failed, the channel will have a non-nil
+	// error value.
+	Ready() <-chan error
```
hwh33: > Ready returns a channel

It seems that this is not quite correct - sometimes `Ready` returns `nil`.  Could you either:
- indicate in this comment that `Ready` can sometimes return `nil`
- or ensure that `Ready` always returns non-nil (perhaps a closed channel?)

chained/water_impl.go
```diff
@@ -23,85 +23,99 @@ type waterImpl struct {
...
 
 	return d, nil
 }
 
+func (d *waterImpl) loadWASM(ctx context.Context, transport string, dir string, wasmAvailableAt string) (io.ReadCloser, error) {
+	locker, ok := waterTransportLocker.Load(transport)
+	if !ok {
+		waterTransportLocker.Store(transport, new(sync.Mutex))
+		locker, _ = waterTransportLocker.Load(transport)
+	}
```
hwh33: This is not concurrency safe.  You can have the following data race:
1. Goroutine 1 calls `Load`, gets `!ok`.
2. Goroutine 2 calls `Load`, gets `!ok`.
3. Goroutine 1 calls `Store` with a new mutex.  Call this mutex M1.
4. Goroutine 1 calls `Load`, gets M1.
5. Goroutine 2 calls `Store` with a new mutex, overwriting the existing entry. Call this mutex M2.
6. Goroutine 2 calls `Load`, gets M2.

Now each goroutine is operating in the code below with a unique mutex.

You _could_ use `sync.Map`'s [CompareAndSwap](https://pkg.go.dev/sync#Map.CompareAndSwap) method to resolve this, but I wouldn't personally recommend that route.  Looking at the documentation for `sync.Map`:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
> 
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

I would follow their recommendation here.  This would look like two variables in the global scope:
```
var (
    // wasmLocks is a map of mutexes protecting RW access to the WebAssembly module files for each
    // WATER transport. A goroutine must acquire the lock for transport "foo" in order to write the
    // foo.wasm file or read the foo.wasm file on disk.
    //
    // Invariant: entries to this map are never overwritten.
    wasmLocks = map[string]*sync.Mutex{}

    // wlLock protects access to the wasmLocks map.
    wlLock = new(sync.Mutex)
)
```

Acquiring the lock for a transport would look like:
```
// Use wlLock to acquire the transport lock. Once acquired, we can let go of wlLock since we
// know the transport lock will never be overwritten.
wlLock.Lock()
m, ok := wasmLocks[transport]
if !ok {
    m = new(sync.Mutex)
    wasmLocks[transport] = m
}
rlLock.Unlock()
// proceed to use m
```

dialer/bandit.go
```diff
@@ -114,6 +114,23 @@ func (o *BanditDialer) chooseDialerForDomain(network, addr string) (ProxyDialer,
 	notAllFailing := hasNotFailing(o.dialers)
 	for i := 0; i < (len(o.dialers) * 2); i++ {
 		d = o.dialers[chosenArm]
+		readyChan := d.Ready()
+		if readyChan != nil {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
hwh33: Since this is critical path for the user, let's just see if the dialer is ready _right now_ rather than wait for a timeout.  So:
```
select {
case err := <-readyChan:
    // Check and handle err
default:
    // Dialer is not ready, call differentArm
}
```

hwh33 (commented):
Mostly looking good now, just a problem with the mutex map and a small note on use of the dialer.

dialer/dialer.go
hwh33: I realize this is nit-picky 🙂 

I just want to make sure future maintainers don't run into unexpected gotchas here!

WendelHime: I've added a comment and also ensured all `IsReady` usages check if the received response is different than nil

dialer/bandit.go
```diff
@@ -114,6 +114,21 @@ func (o *BanditDialer) chooseDialerForDomain(network, addr string) (ProxyDialer,
 	notAllFailing := hasNotFailing(o.dialers)
 	for i := 0; i < (len(o.dialers) * 2); i++ {
 		d = o.dialers[chosenArm]
+		readyChan := d.Ready()
+		if readyChan != nil {
+			select {
+			case err := <-readyChan:
```
hwh33, line 120: Hey sorry, I just realized: this only works once, right?  Once someone reads from the ready channel, there will never again be a value on it.

I think the ready channel should probably only be used to signal readiness and should just be closed when the dialer is ready.  Errors in initialization can be returned by the dial function.

Sorry I didn't catch this earlier!

hwh33 (commented):
Sorry, just noticed one more thing 😬

dialer/bandit.go
WendelHime: > I think the ready channel should probably only be used to signal readiness and should just be closed when the dialer is ready.

Makes sense to me!

> Errors in initialization can be returned by the dial function.

If we do this, the proxy will be punished by the bandit algorithm, should we still do it?

WendelHime: I'll try to create a logic for sending a response for all listeners

WendelHime: Ok, I've updated the ready logic from the water implementation.

Whenever we receive a `ready()` call for checking if the dialer is ready:

1. We check if the dialer is ready, if it is we return a nil response
2. We check if there was an error loading the WASM file.
  2.1 In this case, all other ready calls have already received an answer and they're closed, so this is a late party check
  2.2 In this case we return a temp channel and start a go func that send the error and close the channel.
3. We create a channel, append to a internal list of channels and return to whoever is calling `ready()`.

When we finish loading the WASM by either succeeding or failing, we broadcast the ready status to all channels. If the channel take more than one second for receiving the message, the channel is probably not being listened so it'll be closed.

chained/water_impl.go
```diff
@@ -75,30 +82,95 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			select {
+			case tempChan <- d.errLoadingWASM:
+			case <-ctx.Done():
+			}
+			close(tempChan)
+		}()
+		return tempChan
```
hwh33: I think what you probably want here is a buffered channel:
```
tempChan := make(chan error, 1)
tempChan <- d.errLoadingWASM
return tempChan
```

chained/water_impl.go
```diff
@@ -75,30 +82,95 @@ func newWaterImpl(dir, addr string, pc *config.ProxyConfig, reportDialCore repor
...
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			select {
+			case c <- err:
+			case <-ctx.Done():
+			}
+			close(c)
+		}(err)
+	}
+	wg.Wait()
```
hwh33: Again, buffered channels would greatly simplify this:
```
for _, c := range d.readyChannels {
    c <- err
}
```

dialer/bandit.go
hwh33: > > Errors in initialization can be returned by the dial function.
>
> If we do this, the proxy will be punished by the bandit algorithm, should we still do it?

If there was an error in initialization then the proxy is dead in the water.  In that case, we don't care if it gets punished by the bandit algorithm since it will error every time anyway (actually, we want it to get de-prioritized).

So a simply closed channel, with errors returned on dial attempts would work.  What you've got works too though!  Totally up to you.

hwh33 (approved):
A couple of options for simplification, but this looks good to me!

WendelHime:
I've made an E2E test with this changes and I was able to run WATER smoothly with the newest flashlight changes. The multiplex is also working as expected so this is ready for any other reviews or ready to be merged

hwh33 (approved):
Looks great @WendelHime!  Feel free to merge whenever you're ready

PR #1437, opened 2024-11-07
Two phase dialer implementation that first connects as quickly as possible

myleshorton:
This PR is aimed at reducing the time it takes a user to get connected at startup while also improving reliability. 

The current bandit implementation connects quickly but also experiments with dialers that may not connect at all, thereby hurting the user experience.

This PR switches to a two-phase dialer. In the first phase, we simply try to connect as quickly as possible to all available dialers. If dialers can't connect at all, we discard them. Any dialer that connect at all is immediately available to callers. Then, when we've tested all the dialers for connectivity, we send the connecting dialers and switch to the multi-armed bandit dialer that optimizes for the best dialer to use.


dialer/bandit.go
```diff
@@ -0,0 +1,191 @@
...
+type dataTrackingConn struct {
+	net.Conn
+	dataRecv *atomic.Uint64
+}
+
+func (c *dataTrackingConn) Read(b []byte) (int, error) {
+	n, err := c.Conn.Read(b)
+	c.dataRecv.Add(uint64(n))
+	return n, err
+}
```
myleshorton, line 181: This whole file is just moved to a new package with minor changes

client/client.go
```diff
@@ -354,14 +353,15 @@ func (client *Client) Connect(dialCtx context.Context, downstreamReader io.Reade
...
-func (client *Client) Configure(proxies map[string]*commonconfig.ProxyConfig) []bandit.Dialer {
+func (client *Client) Configure(proxies map[string]*commonconfig.ProxyConfig) []dialer.ProxyDialer {
 	log.Debug("Configure() called")
 	dialers, dialer, err := client.initDialers(proxies)
 	if err != nil {
 		log.Error(err)
 		return nil
 	}
-	client.dialer = dialer
+	client.dialer.set(dialer)
```
myleshorton, line 370: This was actually a bug, as the dialer could be set by multiple goroutines.

myleshorton:
If anyone has a moment to give this a look, I would be very grateful! I think it should help quite a bit with getting users connected quickly.

client/client.go
```diff
@@ -173,14 +173,13 @@ func NewClient(
...
-	}
+
 	client := &Client{
-		configDir:                              configDir,
-		requestTimeout:                         requestTimeout,
-		dialer:                                 banditDialer,
+		configDir:      configDir,
+		requestTimeout: requestTimeout,
+		dialer: &protectedDialer{
+			dialer: dialer.NoDialer(),
```
reflog, line 183: this could use a comment (at least for me this is not very clear)

client/client.go
```diff
@@ -480,57 +465,40 @@ func (client *Client) doDial(
...
 	if client.proxyAll() {
 		log.Tracef("Proxying to %v because proxyall is enabled", addr)
-		op.Set("force_proxied", true)
-		op.Set("force_proxied_reason", "proxyall")
 		return dialProxied(ctx, "whatever", addr)
 	}
 
 	dialDirectForShortcut := func(ctx context.Context, network, addr string, ip net.IP) (net.Conn, error) {
 		log.Debugf("Use shortcut (dial directly) for %v(%v)", addr, ip)
-		op.Set("shortcut_direct", true)
```
reflog: why did you remove all the `op`s?

dialer/fastconnect.go
```diff
@@ -0,0 +1,198 @@
...
+func (fcd *fastConnectDialer) onConnected(pd ProxyDialer, connectTime time.Duration) {
+	log.Debugf("Connected to %v", pd.Name())
+	fcd.connectedLock.Lock()
+	defer fcd.connectedLock.Unlock()
+
+	fcd.connected = append(fcd.connected, connectTimeProxyDialer{
+		ProxyDialer: pd,
+		connectTime: connectTime,
+	})
+	sort.Sort(fcd.connected)
```
reflog: just FYI, there's https://pkg.go.dev/slices#SortFunc if you want to avoid overriding swap/less methods and introducing new type

reflog (approved):
looks good, and I appreciate the move to a more generic ProxyDialer interface, I wasn't a fan of having the word "bandit" all over the codebase :)

client/client.go
myleshorton: Mostly they were annoying me, and I don't think they're being used, but I'll put 'em back =).

dialer/fastconnect.go
myleshorton: I like it! Was poking around for something better but never found that.

PR #1438, opened 2024-11-12
allow blocking otel-reconfiguration

reflog:
This PR adds `ConfigureOnce` to otel package, that will configure Otel and make sure any later calls to Configure will be a no-op.
This allows us to initialize otel earlier, with custom options and ignore settings that come from Global config

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -25,6 +25,8 @@ var (
 
 	stopper   func()
 	stopperMx sync.Mutex
+
+	stopReconfiguration bool
 )
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


otel/otel.go
```diff
@@ -25,6 +25,8 @@ var (
 
 	stopper   func()
 	stopperMx sync.Mutex
+
+	stopReconfiguration bool
```
myleshorton: Maybe an atomic bool here just in case another goroutine happens to call Configure? Otherwise LGTM

myleshorton:
I just went ahead and made the atomic bool switch!

PR #1445, opened 2024-11-20
Stop old dialers when we get a new proxy config

myleshorton:
@garmr-ulfr identified a bug in our dialer whereby we keep dialing old dialers when we get a new proxy config. This stops 'em.

### DIFF
--- a/dialer/dialer.go
+++ b/dialer/dialer.go
@@ -41,7 +41,7 @@ func New(opts *Options) Dialer {
 }
 
 // NoDialer returns a dialer that does nothing. This is useful during startup
-// until a real dialer is available.
+// when we don't yet have proxies to dial through.
 func NoDialer() Dialer {
 	return &noDialer{}
 }
@@ -58,6 +58,9 @@ func (d *noDialer) DialContext(ctx context.Context, network, addr string) (net.C
 }
 
 func (d *noDialer) Close() {}
+
+// Make sure noDialer implements Dialer
+var _ Dialer = (*noDialer)(nil)
 
 const (
 	// NetworkConnect is a pseudo network name to instruct the dialer to establish
--- a/dialer/dialer.go
+++ b/dialer/dialer.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net"
 	"runtime/debug"
+	"sync/atomic"
 	"time"
 
 	"github.com/getlantern/golog"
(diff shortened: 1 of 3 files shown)


dialer/fastconnect.go
```diff
@@ -99,13 +113,23 @@ func (fcd *fastConnectDialer) connectAll(dialers []ProxyDialer) {
...
+			// Loop until we're connected
+			if len(fcd.connected.dialers) < 2 {
+				fcd.parallelDial(dialers)
+				// Add jitter to avoid thundering herd
+				time.Sleep(time.Duration(rand.Intn(4000)) * time.Millisecond)
+			} else {
+				break outerLoop
+			}
+		}
 	}
```
garmr-ulfr: This could be simplified to:
```
for {
	// Loop until we're connected
	if len(fcd.connected.dialers) < 2 {
		fcd.parallelDial(dialers)
	} else {
		break
	}
	select {
	case <-fcd.stopCh:
		log.Debug("Stopping parallel dialing")
		return
	case time.After(time.Duration(rand.Intn(4000)) * time.Millisecond):
	}
}
```
But your way works too!

dialer/dialer.go
```diff
@@ -11,28 +11,37 @@ import (
...
 
 var log = golog.LoggerFor("dialer")
 
+var currentDialer atomic.Value
+
 // New creates a new dialer that first tries to connect as quickly as possilbe while also
 // optimizing for the fastest dialer.
 func New(opts *Options) Dialer {
-	return NewTwoPhaseDialer(opts, func(opts *Options, existing Dialer) Dialer {
+	if currentDialer.Load() != nil {
```
garmr-ulfr, line 27: Are we assuming that there will only ever be one place `New` is called? (ignoring tests)

dialer/fastconnect.go
```diff
@@ -76,6 +83,13 @@ func (fcd *fastConnectDialer) DialContext(ctx context.Context, network, addr str
 	return conn, err
 }
 
+func (fcd *fastConnectDialer) Close() {
+	// We don't call Stop on the Dialers themselves here because they are likely
+	// in use by other Dialers, such as the BanditDialer.
+	// Stop all dialing
+	fcd.stopCh <- struct{}{}
```
garmr-ulfr, line 90: I would make `stopCh` buffered or just close it so this doesn't block until `outerLoop` tries to read from it again.

dialer/fastconnect.go
myleshorton: I like it! Will change.

dialer/dialer.go
myleshorton: It's only called when we get new configs

dialer/fastconnect.go
```diff
@@ -113,21 +113,21 @@ func (fcd *fastConnectDialer) connectAll(dialers []ProxyDialer) {
...
-			// Loop until we're connected
-			if len(fcd.connected.dialers) < 2 {
-				fcd.parallelDial(dialers)
-				// Add jitter to avoid thundering herd
-				time.Sleep(time.Duration(rand.Intn(4000)) * time.Millisecond)
-			} else {
-				break outerLoop
-			}
+
 		}
```
garmr-ulfr: It's not big deal since the goroutine running `connectAll` would only hang around for a few seconds, at most, but, if you add another `case` and use `time.After` instead of `time.Sleep` here, the goroutine will exit as soon as it's canceled and be cleaned up. Just FYI.

garmr-ulfr (approved):
LGTM!

dialer/fastconnect.go
myleshorton: Ah good point -- fixed!

PR #1449, opened 2024-11-21
MAB persist decisions across runs

WendelHime:


dialer/dialer.go
```diff
@@ -82,6 +82,18 @@ type Options struct {
 
 	// OnSuccess is the callback that is called by dialer after a successful dial.
 	OnSuccess func(ProxyDialer)
+
+	// LoadLastBanditRewards is a function that returns the last bandit rewards
+	// for each dialer. If this is set, the bandit will be initialized with the
+	// last metrics.
+	LoadLastBanditRewards func() map[string]BanditMetrics
+
+	SaveBanditRewards func(map[string]BanditMetrics)
```
garmr-ulfr: Do these have to be callbacks? I don't see a need for these be options, or even injected callbacks. It just over complicates things. I would just remove them as options and just always call them.

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+			}
+			metrics[parts[0]] = dialer.BanditMetrics{
+				Reward: reward,
+				Count:  count,
+			}
+		}
+		return metrics
+	}
+}
+
```
garmr-ulfr: We should move these to `bandit.go` in `dialer`. There isn't anything here that shouldn't be accessed by `dialer` and we could pass `configDir` as an option to `dialer`.

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+func loadLastBanditRewards(outputDir string) func() map[string]dialer.BanditMetrics {
+	return func() map[string]dialer.BanditMetrics {
+		dir := filepath.Join(outputDir, "bandit")
+		file := filepath.Join(dir, "rewards.csv")
+		if _, err := os.Stat(file); os.IsNotExist(err) {
+			return nil
+		}
```
garmr-ulfr: `os.ReadFile` will return `ErrNotExist` if the file doesn't exist. This could jus be moved into the `err != nil` check below, then there would be no need to call `os.Stat`.

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+		data, err := os.ReadFile(file)
+		if err != nil {
+			log.Errorf("unable to read bandit rewards from file: %v", err)
+			return nil
+		}
+		lines := strings.Split(string(data), "\n")
+		metrics := make(map[string]dialer.BanditMetrics)
+		for i, line := range lines {
```
garmr-ulfr: Just a suggestion, you could use [bufio.Scanner](https://pkg.go.dev/bufio#Scanner) instead of reading in the entire file and splitting into lines. 
```
f, err := os.Open(file)
if err != nil {
	doStuff
}
scanner := bufio.NewScanner(f)
scanner.Scan()
if err := scanner.Err(); err != nil {
	handleErr
}
header := scanner.Text()
for scanner.Scan() {
	line := scanner.Text()
	doStuff
}
if err := scanner.Err(); err != nil {
	handleErr
}
```

Then it will only read small chunks at a time and, by default, splits on \n.  You way works too!

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+			if i == 0 {
+				continue
+			}
```
garmr-ulfr: Wouldn't an empty line mean it's the last line since that's the only line that should be empty?

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+			reward, err := strconv.ParseFloat(parts[1], 64)
+			if err != nil {
+				log.Errorf("unable to parse reward from %s: %v", parts[1], err)
+				continue
+			}
+			count, err := strconv.Atoi(parts[2])
+			if err != nil {
+				log.Errorf("unable to parse count from %s: %v", parts[2], err)
+				continue
+			}
```
garmr-ulfr: I'm not sure if we would want to continue if there's an issue parsing a line. This file would only exist (and be accurate) if we wrote it, which would mean either the file is corrupted or tampered with and we can't know what's accurate.

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+			log.Errorf("unable to create bandit directory: %v", err)
+			return
+		}
+		file := filepath.Join(dir, "rewards.csv")
+		csv := new(strings.Builder)
+		csv.WriteString("dialer,reward,count\n")
+		for dialerName, metric := range metrics {
+			csv.WriteString(fmt.Sprintf("%s,%f,%d\n", dialerName, metric.Reward, metric.Count))
+		}
+		f, err := os.Create(file)
```
garmr-ulfr: This is where it gets tricky. `os.Create` overwrites the existing file and, currently, you're only writing the rewards for the current dialers. What if, while lantern is running, we get re-assigned a proxy that we had maybe a couple of hours ago? This would treat it as if we've never seen it before. That being said, we may have already decided to ignore that situation since it's uncommon (or maybe will never happen? I don't remember).

client/client.go
garmr-ulfr: Also, there's a [CSV reader](https://pkg.go.dev/encoding/csv#Reader) in the stdlib.

client/client_test.go
```diff
@@ -422,6 +424,89 @@ func TestAccessingProxyPort(t *testing.T) {
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			tempDir, err := os.MkdirTemp("", "client_test")
+			require.NoError(t, err)
+			defer os.RemoveAll(tempDir)
+
+			if err := os.MkdirAll(filepath.Join(tempDir, "bandit"), 0755); err != nil {
+				log.Errorf("unable to create bandit directory: %v", err)
+				return
+			}
```
garmr-ulfr: This can just be consolidated to `require.NoError(t, os.MkdirAll(filepath.Join(tempDir, "bandit"), 0755), "unable to create bandit directory")`

dialer/bandit.go
```diff
@@ -29,21 +29,41 @@ func NewBandit(opts *Options) (Dialer, error) {
...
+			if banditMetrics, ok := dialerWeights[dialer.Name()]; ok {
+				rewards[arm] = banditMetrics.Reward
+				counts[arm] = banditMetrics.Count
+			}
+		}
+		b, err = bandit.NewEpsilonGreedy(0.1, counts, rewards)
+		if err != nil {
+			log.Errorf("unable to create weighted bandit: %w", err)
+			return nil, err
+		}
```
garmr-ulfr, line 73: Do we not need to also call `b.Init` here?

dialer/bandit_test.go
```diff
@@ -79,39 +80,63 @@ func TestBanditDialer_chooseDialerForDomain(t *testing.T) {
...
+		},
+		{
+			name: "should load the last bandit rewards if they exist",
+			opts: &Options{
+				Dialers: []ProxyDialer{oldDialer, newTcpConnDialer()},
+				LoadLastBanditRewards: func() map[string]BanditMetrics {
+					return map[string]BanditMetrics{
+						oldDialer.Name(): {
+							Reward: 0.7,
+							Count:  10,
```
garmr-ulfr: We shouldn't hard-code values like this that we are also testing against. It makes the test fragile and not as easy to maintain. If anyone changes these they would have to make sure they change it everywhere else, and sometimes it's not obvious which values will be affected. This is one of the reasons why a lot of tests use `wantX` args. It's also more readable.

It might also be a good idea to test with multiple dialers to make sure the metrics are assigned to the correct dialer.

dialer/bandit_test.go
```diff
@@ -79,39 +80,63 @@ func TestBanditDialer_chooseDialerForDomain(t *testing.T) {
...
+			assert: func(t *testing.T, got Dialer, err error) {
+				assert.NotNil(t, got)
+				assert.NoError(t, err)
+				assert.IsType(t, &BanditDialer{}, got)
+				rewards := got.(*BanditDialer).bandit.GetRewards()
+				counts := got.(*BanditDialer).bandit.GetCounts()
+				assert.Equal(t, 0.7, rewards[0])
+				assert.Equal(t, 10, counts[0])
+				assert.Equal(t, float64(0), rewards[1])
+				assert.Equal(t, 0, counts[1])
```
garmr-ulfr, line 146: Why is there a second check if there is only only dialer metrics loaded?

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
...
+			return
+		}
+		defer f.Close()
+		if _, err := f.WriteString(csv.String()); err != nil {
+			log.Errorf("unable to write bandit rewards to file: %v", err)
+		}
+	}
+}
+
+func loadLastBanditRewards(outputDir string) func() map[string]dialer.BanditMetrics {
```
garmr-ulfr: Same as above

client/client.go
```diff
@@ -735,10 +737,78 @@ func (client *Client) initDialers(proxies map[string]*commonconfig.ProxyConfig)
 				)
 			}
 		},
+		SaveBanditRewards:     saveBanditRewards(configDir),
+		LoadLastBanditRewards: loadLastBanditRewards(configDir),
 	})
 	return dialers, dialer, nil
 }
 
+func saveBanditRewards(dir string) func(map[string]dialer.BanditMetrics) {
```
garmr-ulfr: We should return any error we encounter and let the caller decided how/if to handle them.

garmr-ulfr (commented):
Nice! Just a few comments.

client/client.go
WendelHime: This line is for skipping CSV headers

dialer/bandit.go
WendelHime: not required, `b.Init` for epsilon greedy just initializes Rewards and Counts as empty fields, in this case we're already doing that and also we're loading the previous rewards too.

dialer/bandit_test.go
WendelHime: This test is doing what you've suggested earlier 😂 , it build a bandit dialer with two dialers, one is the oldest and another is a random new one. In this case `LoadLastBanditRewards` is returning only data from the oldest dialer and the bandit should return the loaded data for this dialer, and for the next dialer, without info, it should have empty reward and count.

client/client.go
WendelHime: Thank you for pointing this, I've just dded a commit for loading the previous rewards before adding new ones, so now we have a reward history per proxy.

dialer/bandit.go
WendelHime: BTW, empty fields based on the given number of arms.

WendelHime:
@garmr-ulfr Thank you for the suggestions! I've made some changes and this PR should be good to be reviewed again!

WendelHime:
Marking this as a draft again, I'll add a feature for ignoring proxies not used in the last 7 days (and stop recording them)

WendelHime:
Now this is ready for review

dialer/dialer.go
```diff
@@ -82,6 +82,15 @@ type Options struct {
...
+
+	// BanditDir is the directory where the bandit will store its data
+	BanditDir string
+}
+
+type banditMetrics struct {
+	Reward    float64
+	Count     int
+	UpdatedAt int64
 }
```
garmr-ulfr: We can move this to `bandit.go`.

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
 	bd.opts.OnSuccess(d)
 	return dt, err
 }
 
+const unusedBanditDialerIgnoredAfter = 7 * 24 * time.Hour
+
+// LoadLastBanditRewards is a function that returns the last bandit rewards
+// for each dialer. If this is set, the bandit will be initialized with the
+// last metrics.
+func (o *BanditDialer) LoadLastBanditRewards() (map[string]banditMetrics, error) {
```
garmr-ulfr: I don't think this needs to be exported, right?

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
+		metrics[line[0]] = banditMetrics{
+			Reward:    reward,
+			Count:     count,
+			UpdatedAt: updatedAt,
+		}
+	}
+	return metrics, nil
+}
+
+func (o *BanditDialer) UpdateBanditRewards(newRewards map[string]banditMetrics) error {
```
garmr-ulfr: I don't think this needs to be exported either.

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, log.Errorf("unable to read line from bandit rewards csv: %w", err)
+		}
+
+		if len(line) != 4 {
+			return nil, log.Errorf("invalid line in bandit rewards csv: %w", line)
+		}
```
garmr-ulfr: You don't need to check if all the fields are there. [Read](https://pkg.go.dev/encoding/csv#Reader.Read) will return `ErrFieldCount` if a line doesn't have the same number of fields as the header.

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
+	}
+
+	file := filepath.Join(o.opts.BanditDir, "rewards.csv")
+	data, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+
+	reader := csv.NewReader(data)
+	_, err = reader.Read() // Skip the header
```
garmr-ulfr: Wow..I can't believe they don't automatically handle the header.

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
+
+	if err = w.Write(headers); err != nil {
+		return log.Errorf("unable to write headers to bandit rewards file: %v", err)
+	}
+
+	for dialerName, metric := range previousRewards {
+		if err = w.Write([]string{dialerName, fmt.Sprintf("%f", metric.Reward), fmt.Sprintf("%d", metric.Count), fmt.Sprintf("%d", metric.UpdatedAt)}); err != nil {
+			return log.Errorf("unable to write bandit rewards to file: %v", err)
+		}
+	}
```
garmr-ulfr, line 282: I thought they would have also handled converting values to strings when writing. It seems like it's easier to just write it yourself since you already have to format it. The `csv.Writer` doesn't seem useful at all..

dialer/bandit.go
```diff
@@ -97,15 +132,147 @@ func (bd *BanditDialer) DialContext(ctx context.Context, network, addr string) (
...
+		reward, err := strconv.ParseFloat(line[1], 64)
+		if err != nil {
+			return nil, log.Errorf("unable to parse reward from %s: %w", line[0], err)
+		}
+		count, err := strconv.Atoi(line[2])
+		if err != nil {
+			return nil, log.Errorf("unable to parse count from %s: %w", line[0], err)
+		}
+
+		metrics[line[0]] = banditMetrics{
```
garmr-ulfr: Would you add header consts and use those for the indexes so it's easy to see what's being referenced?
Like `nameHdr` or something and `line[nameHdr]`. Next to `unusedBanditDialerIgnoredAfter` is fine.

dialer/bandit_test.go
garmr-ulfr: Oh, I see now! There's an old and a new dialer: `Dialers: []ProxyDialer{oldDialer, newTcpConnDialer()},`

dialer/bandit_test.go
```diff
@@ -281,11 +326,112 @@ func Test_differentArm(t *testing.T) {
...
+					Reward: 1.0,
+					Count:  1,
+				},
+			},
+			assert: func(t *testing.T, dir string, err error) {
+				assert.NoError(t, err)
+				f, err := os.Open(filepath.Join(dir, "rewards.csv"))
+				require.NoError(t, err)
+				defer f.Close()
+				b, err := io.ReadAll(f)
```
garmr-ulfr, line 348: This could all be condensed to `b, err := os.ReadFile(filepath.Join(dir, "rewards.csv"))`. :smiley:

dialer/bandit_test.go
```diff
@@ -281,11 +326,112 @@ func Test_differentArm(t *testing.T) {
...
+				require.NoError(t, err)
+
+				lines := strings.Split(string(b), "\n")
+				// check if headers are there
+				assert.Contains(t, lines[0], "dialer,reward,count,updated at")
+				// check if the data is there
+				assert.Contains(t, lines[1], "test-dialer,1.000000,1")
```
garmr-ulfr: Shouldn't these be `Equal` instead of `Contains`?

dialer/bandit_test.go
```diff
@@ -281,11 +326,112 @@ func Test_differentArm(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			tempDir, err := os.MkdirTemp("", "client_test")
+			require.NoError(t, err)
+			defer os.RemoveAll(tempDir)
+			require.NoError(t, os.MkdirAll(filepath.Join(tempDir, "bandit"), 0755), "unable to create bandit directory")
+
+			f, err := os.Create(filepath.Join(tempDir, "rewards.csv"))
+			require.NoError(t, err)
+			defer f.Close()
+			_, err = f.WriteString(tt.given)
```
garmr-ulfr: This could all be condensed to 
`err = os.WriteFile(filepath.Join(tempDir, "rewards.csv"), []byte(tt.given), 0644)`. :smiley:

dialer/bandit_test.go
```diff
@@ -281,11 +326,112 @@ func Test_differentArm(t *testing.T) {
...
+				assert.Empty(t, metrics)
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tempDir, err := os.MkdirTemp("", "client_test")
+			require.NoError(t, err)
+			defer os.RemoveAll(tempDir)
+			require.NoError(t, os.MkdirAll(filepath.Join(tempDir, "bandit"), 0755), "unable to create bandit directory")
```
garmr-ulfr: Did you mean to use `tempDir/bandit` as `BanditDir`?

garmr-ulfr (commented):
Much better! Just a few more comments.

dialer/bandit.go
WendelHime: uuuuuu that would be nice, I'll do that

dialer/bandit_test.go
WendelHime: I guess I was using contains before because I was using the whole CSV as string, I'll replace it

garmr-ulfr (approved):
LGTM!

PR #1452, opened 2024-11-25
Initialize all domain fronting code one time on startup

myleshorton:
The current startup sequence suffers from some byzantine code for initializing the stored global config (either embedded or saved), which is then required for initializing all fronting code that is necessary for config fetches, particularly for first time users. This is exacerbated by the getlantern/fronted code, which is very rapidly initialized twice in many cases -- once with using the global config on disk, and once using the fetched global config. In that process, any previously verified and working fronts are discarded, but also any code unlucky enough to be called prior to the new global config being fetched keeps mindlessly chugging through outdated fronts.

This attempts to address all that by initializing the fronting code right away in a straightforward manner and then updating the fronting code with new fronts from the network vs replacing it wholesale.


proxied/fronted.go
```diff
@@ -35,10 +26,5 @@ func (f frontedRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
 		op := ops.Begin(f.opName)
 		defer op.End()
 	}
-	rt, ok := fronted.NewFronted(f.masqueradeTimeout)
-	if !ok {
-		return nil, errors.New("Unable to obtain direct fronter")
-	}
-	changeUserAgent(req)
-	return rt.RoundTrip(req)
+	return fronter.RoundTrip(req)
```
myleshorton, line 29: The fronter now handles the timeout internally based on the request context

myleshorton:
I'm actually going to pull this in...it looks much bigger than it is, but it's all just to use the new fronting code.

PR #1453, opened 2024-11-26
VMess Support

reflog:
complements https://github.com/getlantern/http-proxy/pull/632


chained/vmess_impl.go
```diff
@@ -0,0 +1,100 @@
...
+// newVmessImpl creates a new VMess proxy implementation.
+// supports the following options:
+// - uuid: the UUID of the user on the VMess server
+// - security: the security level to use, defaults to "auto".  Options are:
+//  "auto": automatically determine the security level
+//  "none" or "zero": no security
+//  "aes-128-cfb": legacy security
+//  "aes-128-gcm": AES-128-GCM security
+//  "chacha20-poly1305": ChaCha20-Poly1305 security
+
```
garmr-ulfr: Stray newline.

chained/vmess_impl.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, errors.New("failed to create vmess client: %v", err)
+	}
+
+	var seed int64
+	err = binary.Read(crand.Reader, binary.BigEndian, &seed)
+	if err != nil {
+		return nil, errors.New("unable to initialize rng: %v", err)
+	}
+
+	source := mrand.NewSource(seed)
```
garmr-ulfr: I know this was just copied from shadowsocks_impl but we should start using `math/rand/v2` with any new code (and at some point update the existing code to use it). These lines can be removed as v2 handles global seeding on a goroutine basis; meaning we can just use the global `rand.IntN` and don't need to create a new `Rand` instance each time. The global functions are also concurrency safe, so no more need for a mutex.

chained/vmess_impl.go
```diff
@@ -0,0 +1,100 @@
...
+	impl.rngmx.Lock()
+	defer impl.rngmx.Unlock()
+	// [2 - 22]
+	sz := 2 + impl.rng.Intn(21)
+	b := make([]byte, sz)
+	for i := range b {
+		b[i] = letters[impl.rng.Intn(len(letters))]
+	}
+	return string(b) + ".com"
+}
```
garmr-ulfr: Switching to `math/rand/v2`, this would be reduced to: 
```
func (impl *vmessImpl) generateUpstream() string {
	const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
	// [2 - 22]
	sz := 2 + rand.IntN(21)
	b := make([]byte, sz)
	for i := range b {
		b[i] = letters[rand.IntN(len(letters))]
	}
	return string(b) + ".com"
}
```

garmr-ulfr (approved):
LGTM! Just a couple of comments.

reflog:
@garmr-ulfr  done

garmr-ulfr (approved):
LGTM

PR #1455, opened 2024-12-02
Update protobuf package to resolve namespace conflict

atavism:
```
[   33.419548] sh[1096]: DEBUG flashlight.common: const.go:69 ****************************** stagingMode: false
[   33.419808] sh[1096]: DEBUG flashlight.common: version.go:28 Flashlight version is 7.6.147-0.20241130213647-72d9979e5a3e
[   33.419864] sh[1096]: DEBUG flashlight.common: version.go:31 Setting LibraryVersion to 7.6.147-0.20241130213647-72d9979e5a3e
[   33.421989] sh[1096]: panic: proto: file "types.proto" is already registered
[   33.422156] sh[1096]: 		previously from: "github.com/getlantern/lantern-cloud/cmd/api/apipb"
[   33.422219] sh[1096]: 		currently from:  "github.com/getlantern/flashlight/v7/issue"
[   33.422286] sh[1096]: 	See https://protobuf.dev/reference/go/faq#namespace-conflict
```


issue/types.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package issue;
+package flashlight.issue;
```
atavism: I'm not sure if this change is needed as well @garmr-ulfr @myleshorton

atavism:
Looks like this was already fixed? Happy to close this if need be

myleshorton:
hmnn...I don't see any recent commits

issue/types.proto
```diff
@@ -1,5 +1,4 @@
 syntax = "proto3";
-package issue;
```
atavism, line 2: I think we can just remove this anyway

issue/types.proto
garmr-ulfr: I think this just adds a scope to the definitions.  I'm not sure though.

myleshorton:
I think this ended up just going away with changes to the init sequence -- basically calling into the issue package early caused the conflict, whereas normally pinger would never have any reason to call the issue stuff at all.

PR #1459, opened 2024-12-04
Upgrade utls to v1.6.7

Jovis7:
* In https://github.com/getlantern/engineering/issues/292, it's found that `http-proxy` needs to upgrade utls to fix the bug.
* We should also update utls across the rest of our code base for good measure : https://github.com/getlantern/engineering/issues/1757
* go.mod : [http-proxy](https://github.com/getlantern/http-proxy/pull/633) / [tlsdialer](https://github.com/getlantern/tlsdialer/pull/25) / [tlsresumption](https://github.com/getlantern/tlsresumption/pull/5) / [fronted](https://github.com/getlantern/fronted/pull/52) also depend on utls, so flashlight also update dependency to these modules.
* For utls 1.6.7 (that depends on newer Go), it needs to remove `RSA based cipher suites`.
https://go.dev/doc/go1.22 : By default, cipher suites without ECDHE support are no longer offered by either clients or servers during pre-TLS 1.3 handshakes. This change can be reverted with the tlsrsakex=1 GODEBUG setting.
* With new utls, the way (`UConn.HandshakeState.Session`) we use to get `utls.ClientSessionState` doesn't work anymore.
   New utls uses `utls.ClientSessionCache.Get(host)` to get `utls.ClientSessionState`.
* [As of Go 1.16, the ioutil package is deprecated.](https://pkg.go.dev/io/ioutil)
* Sync new field `session_state` in `apidb/types.proto` with [lantern-cloud](https://github.com/getlantern/lantern-cloud/pull/1164/files)


WendelHime (approved):
LGTM! I've also executed some E2E tests and was able to connect and browse with latest lantern-client changes

PR #1460, opened 2024-12-08
#1757 Delete old tls state file

Jovis7:
* For [upgrading new utls(v1.6.7)](https://github.com/getlantern/engineering/issues/1757) on lantern-client, some old fields (Vers / CipherSuite / MasterSecret) of ClientSessionState were removed from [serializedClientSessionState struct](https://github.com/getlantern/tlsresumption/pull/5/commits/ae315acc66b6bd8f5202be77f32f95f76c1eaf43#).
This is intended as new utls doesn't work with old fields.
* However, when old client app upgraded to new client app, the first time it would fail to parse old tls state file on disk.  To handle this, LibVersion (e.g. 7.6.150) is added to the beginning of state file.   State file should be deleted if it's version is different from current.
* The new states will be [saved to file](https://github.com/getlantern/flashlight/blob/1bebe9ba61d23cd2649dab5014862fbc692b97b7/chained/tls_file_cache.go#L89-L116) once it [gets from server](https://github.com/getlantern/flashlight/blob/1bebe9ba61d23cd2649dab5014862fbc692b97b7/chained/tls_file_cache.go#L125-L136).

### DIFF
--- a/chained/tls_file_cache.go
+++ b/chained/tls_file_cache.go
@@ -69,6 +69,7 @@ func persistSessionStates(configDir string, saveInterval time.Duration) {
 	if err == nil {
 		log.Debugf("Initializing current session states from %v", filename)
 		rows := strings.Split(string(existing), "\n")
+		atLeastOneStateParsedSuccessfully := false
 		for _, row := range rows {
 			state, err := parsePersistedState(row)
 			if err != nil {
@@ -76,6 +77,14 @@ func persistSessionStates(configDir string, saveInterval time.Duration) {
 				continue
 			}
 			currentSessionStates[state.server] = *state
+			atLeastOneStateParsedSuccessfully = true
+		}
+		// Delete the state file if none of the states were parsed successfully. This could happen if state struct is upgraded.
+		if !atLeastOneStateParsedSuccessfully {
+			log.Errorf("No valid session states found, deleting %v", filename)
+			if err := os.Remove(filename); err != nil {
+				log.Errorf("Failed to delete %v: %v", filename, err)
+			}
 		}
 	}
 
--- a/chained/tls_file_cache.go
+++ b/chained/tls_file_cache.go
@@ -64,18 +64,14 @@ func PersistSessionStates(configDir string) {
 	})
 }
 
-func isSameLibraryVersion(rows []string) bool {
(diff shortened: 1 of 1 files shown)


garmr-ulfr:
I don't know if we want to delete the file if we fail to parse it. That could mask an issue with writing/parsing the session states, such that it fails to parse it every time, even if with the most recent version. I think we should write the flashlight version to the beginning of the file. Then we can just check if the first line is the version number and, if not, we know it's from a previous version so then we delete it. This was, it's backwards compatible while still being able to catch deeper issues.

Jovis7:
@garmr-ulfr   Thanks for the comment!  I've tried to implemented it (write the flashlight version to the beginning of the file).

chained/tls_file_cache.go
```diff
@@ -62,26 +64,39 @@ func PersistSessionStates(configDir string) {
...
+		log.Debugf("tls_session_states LibraryVersion %v", fileLibraryVersion)
+		log.Debugf("current LibraryVersion %v", common.LibraryVersion)
+		if fileLibraryVersion == common.LibraryVersion {
+			return true
+		}
+	} else {
+		log.Debugf("%v string not found in tls_session_states", libraryVersionString)
+	}
+	return false
+}
```
garmr-ulfr: We only need to check the first line, so we don't need to pass in all rows. Since `rows` is a slice, you can call `isSameLibraryVersion(rows[0])` and it will be set to the empty string. 

We also should check if `fileLibraryVersion >= common.LibraryVersion`, in case a user skips this version and upgrades to the next one from an older version.

garmr-ulfr: This part isn't absolutely necessary, but best practices for go is to return early. So this can be simplified to:
```
func isSameLibraryVersion(line string) bool {
	if !strings.HasPrefix(line, libraryVersionString) {
		return false
	}
	fileLibraryVersion := strings.TrimPrefix(line, libraryVersionString)
	log.Debugf("tls_session_states LibraryVersion %v", fileLibraryVersion)
	return fileLibraryVersion >= common.LibraryVersion 
}
```

garmr-ulfr (approved):
Just one last thing, otherwise it looks good! It's a small change so after you address it, you can go ahead and merge.

PR #1464, opened 2024-12-11
added flashlight-tester, to be used by pinger

reflog:
please see tester/README.md for details

### DIFF
--- a/.github/workflows/buid_tester.yml
+++ b/.github/workflows/buid_tester.yml
@@ -37,8 +37,6 @@ jobs:
           push: true
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          secrets: |
-            "github_token=${{ secrets.CI_PRIVATE_REPOS_GH_TOKEN }}"
           file: ./tester/Dockerfile
           tags: |
             us-docker.pkg.dev/lantern-cloud/containers/flashlight-tester:latest
--- a/.github/workflows/buid_tester.yml
+++ b/.github/workflows/buid_tester.yml
@@ -1,0 +1,45 @@
+name: Build and Push flashlight tester image
+on:
+  push:
+    branches: [main]
+jobs:
+  docker_push:
+    name: docker push
+    runs-on:
+      group: large-runners
+    permissions:
+      contents: "read"
+      id-token: "write"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/buid_tester.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and Push flashlight tester image
+on:
+  push:
+    branches: [main]
```
WendelHime, line 4: Can we update this for running on pull requests (since it's not merged it'll only run on this PR) so we can see if the pipeline is working as expected, then we can revert the commit change

WendelHime: Forget it was already executed and worked

WendelHime: https://github.com/getlantern/flashlight/actions/runs/12279823034

reflog: @WendelHime yeah, I did test it before. I didn't want it to run on EVERY PR though

tester/README.md
```diff
@@ -0,0 +1,48 @@
...
+> Note: this tester does not time out. It will run indefinitely until it's stopped, or it succeeds in pinging the target URL.
+
+## Environment variables
+
+- `DEVICE_ID`: The device id to use.
+- `USER_ID`: The user id to use.
+- `TOKEN`: The token to use.
+- `RUN_ID`: The run id to use. It will be added to honeycomb traces as value of attribute `pinger-id`. (you can use it for looking up traces for the specific run)
+- `TARGET_URL`: The target url to use. This is the url that will be pinged through flashlight.
+- `OUTPUT`: The output path to use. This is the path where the output will be written to (proxies.yaml, global.yaml, etc). You can place proxies.yaml there to use it instead of fetching.
```
hwh33: There's a lot of language in here about the "output directory" which in reality appears to be the config directory.  Why not just call it the config directory?  That's what the team is used to calling that folder.  Given that it holds config, it's also more of an input directory which adds some confusion to the new lingo IMO.

hwh33: I see that you _are_ writing output files to the config directory.  You could either (a) define a new directory for that, accepting both an `OUTPUT_DIR` and `CONFIG_DIR` (which could be the same) or (b) just state that the program will write output files to the config directory.

hwh33 (approved):
A minor point of confusion over naming (feel free to ignore), but this looks good.

PR #1466, opened 2024-12-12
updating algorithm to calculate speed connection for MAB

WendelHime:
This PR is adding a field for storing the elapsed time to `Read` a given dialer connection and use this value to calculate the speed connection instead of using the time until we sample the data. I've also made some other changes so we can be able to test operations that rely on `time.AfterFunc`.


hwh33 (approved):
Nice work @WendelHime!  This is a good improvement over what we've got.

This is still limited by the upstream the client happens to be connecting to.  In the future we could consider throughput and latency tests directly against the proxies themselves.  But maybe that's overkill!

Anyway, feel free to merge.

WendelHime:
> This is still limited by the upstream the client happens to be connecting to. In the future we could consider throughput and latency tests directly against the proxies themselves. But maybe that's overkill!

🤔 I don't think that's overkill, this could be an awesome feature for pre-setting the bandit rewards

PR #1467, opened 2024-12-12
Make proxied and fronting code self-contained to avoid platform-specific errors

myleshorton:
We've been having issues with initializing the `proxied` package consistently for all platforms. iOS, for example, does not call `flashlight.New`, and, as a result, the `fronter` responsible for domain fronting code in the `proxied` package was never set. This change is more accommodating of heterogeneous initialization environments with the goal of ensuring fronting works as expected for all users.

### DIFF
--- a/config/client_config.go
+++ b/config/client_config.go
@@ -8,12 +8,6 @@ import (
 	"github.com/getlantern/fronted"
 )
 
-const (
-	// for historical reasons, if a provider is unspecified in a masquerade, it
-	// is treated as a cloudfront masquerade (which was once the only provider)
-	DefaultFrontedProviderID = "cloudfront"
-)
-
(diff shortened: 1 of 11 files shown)


config/client_config.go
```diff
@@ -8,12 +8,6 @@ import (
 	"github.com/getlantern/fronted"
 )
 
-const (
-	// for historical reasons, if a provider is unspecified in a masquerade, it
-	// is treated as a cloudfront masquerade (which was once the only provider)
-	DefaultFrontedProviderID = "cloudfront"
-)
-
```
myleshorton, line 16: This is now contained within the `fronted` repo

myleshorton:
This is mostly test changes -- pulling it in to hopefully fix iOS main

PR #1470, opened 2024-12-13
Revert "MAB persist decisions across runs"

myleshorton:
Reverts getlantern/flashlight#1449


myleshorton:
Reverting for now so we can push stuff out.

PR #1471, opened 2024-12-17
Report Issue fix

jigar-f:
This PR passes the country to report the issue for IOS, this is needed due to report issue using the new geoLookup API for getting the country, and for IOS we are still using the old geo lookup

### DIFF
--- a/issue/issue.go
+++ b/issue/issue.go
@@ -42,6 +42,7 @@ func SendReport(
 	model string, // alphanumeric name
 	osVersion string,
 	attachments []*Attachment,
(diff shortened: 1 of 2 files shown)


garmr-ulfr (approved):
I know this is supposed to be temporary, but I wonder if we should just keep it this way. Or maybe even remove the lookup on "" and just rely on the caller to get the country.

jigar-f:
I agree. We can keep it this way; it looks much cleaner. That being said, for now, Let me merge this. I will revisit this later.

PR #1478, opened 2025-01-16
Revert changes for ads.

jigar-f:
I need to revert this unless 50% of ad requests will fail. We can't push changes until we push a new build to Pord, Completely forgot that these changes will live to the previous build as well, my bad


reflog (approved):

jigar-f:
Thanks @reflog

WendelHime (approved):

PR #1479, opened 2025-01-16
add headers to issue request

thomasjib:
tickets haven't been able to read the flashlight version, and I think it is because that header doesnt appear to get added before the request is sent to lantern-cloud. Should fix https://github.com/getlantern/engineering/issues/1776

### DIFF
--- a/issue/issue.go
+++ b/issue/issue.go
@@ -80,7 +80,6 @@ func sendReport(
 	attachments []*Attachment,
 	country string,
 ) error {
-	log.Debug("HERE")
 	httpClient := &http.Client{
 		Transport: proxied.Fronted("issue_fronted_roundtrip"),
 	}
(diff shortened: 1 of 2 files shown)


jay-418:
@thomasjib I think this looks good, but it would be great if someone from @getlantern/sdk could review

issue/issue.go
```diff
@@ -144,6 +146,10 @@ func sendReport(
 	}
 	req.Header.Set("content-type", "application/x-protobuf")
 
+	common.AddCommonNonUserHeaders(userConfig, req)
+
+	log.Debugf("isssue sendReport X-lantern-version header: %v", req.Header.Get("X-Lantern-Version"))
```
WendelHime:
```suggestion
	log.Debugf("issue sendReport X-lantern-version header: %v", req.Header.Get("X-Lantern-Version"))
```

WendelHime (approved):

PR #1480, opened 2025-01-17
Use kindling instead of proxied

myleshorton:
This is a pretty massive change to use a single kindling HTTP client instead of the proxied package.


myleshorton:
Gonna pull this in so we can get it into beta testing!

PR #1481, opened 2025-01-20
reduced the number of init() calls and references to staging

reflog:
mostly the reason for this cleanup is - our over-use of init() function causes extra logs to leak out before Flashlight is actually initialized

oh, and I removed references to Staging, since... it doesn't exist.

### DIFF
--- a/common/const.go
+++ b/common/const.go
@@ -12,9 +12,6 @@ const (
 	// UserConfigURL is the URL for fetching the per user proxy config.
 	UserConfigURL = "http://df.iantem.io/api/v1/config"
 
-	// UserConfigStagingURL is the URL for fetching the per user proxy config in a staging environment.
-	UserConfigStagingURL = "https://api-staging.iantem.io/config"
-
 	// Sentry Configurations
 	SentryTimeout         = time.Second * 30
 	SentryMaxMessageChars = 8000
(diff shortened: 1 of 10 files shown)


common/const.go
```diff
@@ -33,50 +30,16 @@ var (
...
-	if err != nil {
-		log.Errorf("Error parsing boolean flag: %v", err)
-		return
-	}
-	if Staging {
-		ProAPIHost = "api-staging.getiantem.org"
-	}
-	forceAds, _ = strconv.ParseBool(os.Getenv("FORCEADS"))
-}
+var forceAds, _ = strconv.ParseBool(os.Getenv("FORCEADS"))
```
WendelHime: I guess this can also be removed with the function below, I don't see any reference or usage (it looks like the same file exists in lantern-client `internalsdk/common/const.go` and also is not being used)

WendelHime: probably this whole package shouldn't exist, this definitions are only used at specific packages and most stuff is unused

WendelHime (approved):

common/const.go
myleshorton: Agreed forceads can go

myleshorton (approved):
LGTM!

PR #1482, opened 2025-01-23
add new UPDATE_FAIL issue type

jay-418:
partially addresses
- https://github.com/getlantern/engineering/issues/1823


issue/types.pb.go
```diff
@@ -91,18 +94,16 @@ func (Request_ISSUE_TYPE) EnumDescriptor() ([]byte, []int) {
...
 }
 
 func (x *Response) Reset() {
 	*x = Response{}
-	if protoimpl.UnsafeEnabled {
-		mi := &file_types_proto_msgTypes[0]
-		ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
-		ms.StoreMessageInfo(mi)
-	}
+	mi := &file_types_proto_msgTypes[0]
```
jay-418, line 104: these unexpected changes suggest I've done something wrong with generating a type update

issue/types.pb.go
```diff
@@ -113,7 +114,7 @@ func (*Response) ProtoMessage() {}
 
 func (x *Response) ProtoReflect() protoreflect.Message {
 	mi := &file_types_proto_msgTypes[0]
-	if protoimpl.UnsafeEnabled && x != nil {
```
jay-418, line 116: UnsafeEnabled?

issue/types.pb.go
hwh33: I tried a small change as well and got the same thing.  I'm guessing something changed in the implementation of protocol buffers in Go between protoc-gen-go v1.30.0 and v1.36.3.  These changes don't look like they'd impact use or serialization of the messages (I can't tell for sure from a quick look, but nothing jumps out).

I think you're good!

issue/types.pb.go
hwh33: As noted above, probably just a change in the implementation of protocol buffers between versions.

hwh33 (approved):
LGTM!

PR #1492, opened 2025-04-14
Dial using proxyless in parallel with proxies

myleshorton:
This first always tries to dial directly using proxyless if we can.


myleshorton:
Gotta fix one quirk with the outline stuff.

dialer/proxyless.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		host = addr
+	}
+	domains := []string{host}
+
+	configBytes, err := embedFS.ReadFile("smart_dialer_config.yml")
+	if err != nil {
+		log.Errorf("Failed to read smart dialer config: %v", err)
+		return nil, err
+	}
```
WendelHime: We probably want to load this file once instead of reading every dial, so we could move this to `newProxylessDialer` and add a attribute for config on `proxylessDialer`

WendelHime: I guess it would be interesting to have a map of proxyless techniques per country and maybe provide them at the global config, but for now this is good

dialer/proxyless.go
```diff
@@ -0,0 +1,117 @@
...
+	var host string
+	var port string
+	var err error
+	if host, port, err = net.SplitHostPort(address); err != nil {
+		addr = fmt.Sprintf("%s:%s", host, "443")
+	} else if port == "" || port == "443" {
+		addr = fmt.Sprintf("%s:%s", host, "443")
+	} else {
+		return nil, fmt.Errorf("proxyless can only dial 443: %s", addr)
+	}
```
garmr-ulfr: It's not really important but, this can all be simplified to:
```
host, port, _ := net.SplitHostPort(address)
if port != "" && port != "443" {
	return nil, fmt.Errorf("proxyless can only dial 443: %s", addr)
}
addr := host + ":443"
```

dialer/proxyless.go
```diff
@@ -0,0 +1,117 @@
...
+	// Check if the dialer already exists in the map
+	if dialer, ok := d.dialers.Load(host); ok {
+		// If it exists, use it
+		return dialer.(transport.StreamDialer).DialStream(ctx, address)
+	}
+
+	domains := []string{host}
+	dialer, err := d.strategyFinder.NewDialer(context.Background(), domains, d.configBytes)
```
garmr-ulfr: Shouldn't we also use the provided ctx here?

dialer/proxyless.go
WendelHime: @myleshorton what do you think about adding the smart dialer config as part of the global config? We could add different techniques later and it wouldn't be hard-coded

WendelHime: I'm afraid that users in the future might stop updating on this version (by any random reason which happens) and the techniques implemented here would be frozen

garmr-ulfr (approved):
Just a couple of comments, but otherwise LGTM!

dialer/proxyless.go
myleshorton: yes!

myleshorton:
I'm taking another pass at this to parallelize it...will have it in tomorrow!

.github/workflows/go.yml
```diff
@@ -17,17 +17,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: "go.mod"
-      - name: Install go go-ctrf-json-reporter
-        run: go install github.com/ctrf-io/go-ctrf-json-reporter/cmd/go-ctrf-json-reporter@latest
```
myleshorton, line 21: They updated this repo/GH action, so this started failing.

myleshorton:
Hey @WendelHime and @garmr-ulfr I ended up making a bunch of changes here to try proxyless in parallel to make sure this only improves upon the current situation for users. I ended up giving the proxyless dialers their own context because otherwise the context from the proxied dialer would cancel the proxyless whenever it succeeded first.

client/client_test.go
```diff
@@ -149,7 +149,7 @@ func TestServeHTTPTimeout(t *testing.T) {
 	req, err := http.NewRequest("CONNECT", "https://a.com:443", nil)
 	require.NoError(t, err)
 	resp, err := roundTrip(client, req)
-	require.Contains(t, err.Error(), "deadline")
+	require.Contains(t, err.Error(), "any dialers")
```
garmr-ulfr: Is error non-nil on success?

garmr-ulfr:
Reviewing now

myleshorton:
I just realized I think I lost track of the state change in there -- i.e. FastConnect has the logic to switch to the Bandit, but Bandit doesn't actually get used I don't think! Will fix!

dialer/parallel_dialer.go
```diff
@@ -0,0 +1,77 @@
...
+	var connCh = make(chan net.Conn, 2)
+	var errCh = make(chan error, 2)
+	dialers := []Dialer{d.proxylessDialer, d.dialer}
+	for _, d := range dialers {
+		go func(d Dialer) {
+			conn, err := d.DialContext(ctx, network, addr)
+			if err != nil {
+				errCh <- err
+			}
+			connCh <- conn
```
garmr-ulfr: This should be wrapped in an `else`. Currently, it still sends a value (`nil`) on error and so the following select statement will sometimes read from `connCh`, instead of `errCh`, returning `nil,nil`.

myleshorton: oooh good catch!!

PR #1510, opened 2025-07-31
add tracks to issue submission

thomasjib:
adds the tracks used to issue submissions, so we can see which tracks are causing problems.


issue/issue.go
```diff
@@ -129,6 +131,22 @@ func sendReport(
 		}
 	}
 
+	// get which tracks the user was given in their config, if any are returned in less than 2 seconds
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+	defer cancel()
```
garmr-ulfr: You don't really need to wait here; either there are or there aren't tracks when the issue is being submitted. If there aren't, this will wait for the context to expire. This could be changed to:
```go
ctx, _ := context.WithTimeout(context.Background(), -time.Second)
```
so it will return immediately.

garmr-ulfr (approved):
LGTM! Just the one comment

PR #1512, opened 2025-11-28
fix exponential backoff overflow

magras:
The `backoff` function's wait time calculation completely overflows `time.Duration` on the 55th retry (approximately after 6 hours). This results in zero wait times, leading to the uncontrolled spawn of hundreds of goroutines, which can cause memory exhaustion and OOM kill on linux.

### DIFF
--- a/services/http.go
+++ b/services/http.go
@@ -16,7 +16,7 @@ const (
 )
 
 // sender is a helper for sending post requests. If the request fails, sender calculates an
-// exponential backoff time using retryWaitSeconds and return it as the sleep time.
+// exponential backoff time using minRetryWait and return it as the sleep time.
 type sender struct {
 	nextWait time.Duration
 }
--- a/services/http.go
+++ b/services/http.go
@@ -2,7 +2,6 @@ package services
 
 import (
 	"fmt"
-	"math"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


magras:
There is a test program that demonstrates the problem:
```
package main

import (
	"fmt"
	"math"
	"time"
)

func main() {
	n := 60
	for i := 0; i < n; i++ {
		x := math.Pow(2, float64(i))
		d := time.Duration(x) * time.Second
		fmt.Println(i, x, d)
	}
}
```
playground: https://go.dev/play/p/wBxP0SpRRkt

magras:
Hey @hwh33, @myleshorton, @reflog, could one of you please take a look at this PR?

reflog:
@magras I'm on it, thank you for reporting

reflog (changes requested):
There may be a subtle bug: if the first request fails before any success, nextWait starts at zero (Go's default for time.Duration):

  s.nextWait = retryWaitSeconds  // only set on success

  On first failure → wait = 0 → immediate retry with no backoff.

  Recommendation: Initialize nextWait to retryWaitSeconds in the struct or handle the zero case in backoff():

  if s.nextWait == 0 {
      s.nextWait = retryWaitSeconds
  }

magras:
> Recommendation: Initialize nextWait to retryWaitSeconds in the struct

I don't know go, but I found no native way to achieve that.

Hence I went the second way, but slightly deviated from it: now retry timeout is clamped between min and max.

